PR #4, opened 2016-12-06
Additional fields

bibliotechy:
Adds indexing for locations and control numbers.

Adds facet for item Locations

Merges display fields that currently have multiple values into single values.

Adds mapping of location codes to location names.

Updates the solr schema to allow format to be multivalued.


app/controllers/catalog_controller.rb
```diff
@@ -117,6 +121,8 @@ class CatalogController < ApplicationController
     config.add_show_field 'published_vern_display', label: 'Published'
     config.add_show_field 'lc_callnum_display', label: 'Call number'
     config.add_show_field 'isbn_t', label: 'ISBN'
+    config.add_show_field 'control_number_display', label: 'Control Number'
+    onfig.add_show_field 'location_display', label: 'Location', helper_method: :render_location
```
skng5: Typo here. Dropped a 'c'

bibliotechy:
@skng5 thanks for catching that. Now fixed.

PR #8, opened 2017-01-18
Integrate with Alma Ruby API

skng5:
Issue #COB-81

- Updated method calls
- Temporary instructions to add Alma_ID to users


bibliotechy:
:+1:

PR #15, opened 2017-02-07
Include the Blacklight Alma Gem

bibliotechy:
Includes a fork of the Blacklight Alma gem into our app.

Includes marc indexer that can be used with traject cli.


skng5:
Looks good.

PR #16, opened 2017-02-07
Remove import of nonexistent scss file.

bibliotechy:
Make application stylesheet scss.

@cdoyle-temple Was working fine in development, but when we switched to `RAILS_ENV=production` for deployments, the `@import` of a file that was not preent caused issues.


skng5:
👍

PR #30, opened 2017-05-19
Optional, configurable relative_url_root

bibliotechy:
Sets `config.relative_url_root` if the config file exists.


skng5:
Looks good 👍

PR #33, opened 2017-06-30
Feature specs

skng5:
@bibliotechy I'd like to make a PR so we can address subsequent in separate branches.


spec/features/indices_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+        click_link item['title']
+        expect(current_url).to eq item_url
+        within("h3") do
+          expect(page).to have_text item['title']
+        end
+        click_link item['title']
+      end
+    end
+
+    scenario "User visits a document directoy" do
```
bibliotechy: typo

skng5:
@bibliotechy Missed one - Leftover Pry

PR #37, opened 2017-07-06
133 error messages user friendly and targeted

sensei100:


app/controllers/errors_controller.rb
```diff
@@ -0,0 +1,9 @@
+class ErrorsController < ApplicationController
+  def not_found
+    render(:status => 404)
+  end
+
+  def internal_erver_error
```
bibliotechy: @sensei100 looks like an "s" was missed both here and in the view file name

config/alma.yml.example
```diff
@@ -1,5 +1,5 @@
 default: &default
-  apikey: "EX_LIBRIS_DEVELOPERS_NETWORK_APIKEY"
+  apikey: "l7xxb45c21271ab94ddfb7fcb7221e5b2ef6"
```
bibliotechy: OK, this api key should not be committed in the example. It should only have been changed in the `alma.yml` you copied this file to, not in the example.

sensei100:
Those things should be fixed now.

config/routes.rb
```diff
@@ -1,5 +1,9 @@
 Rails.application.routes.draw do
 
+  get 'errors/not_found'
+
+  get 'errors/internal_erver_error'
```
bibliotechy: I don't think we need these get routes in addition to the `match '/404'` and `match '/500'` below.

sensei100:
I removed the get routes for errors.

PR #55, opened 2017-07-28
BL-26 Add link from creator_display_field

sensei100:
Creator links to search results


app/controllers/catalog_controller.rb
```diff
@@ -25,6 +25,7 @@ class CatalogController < ApplicationController
 
     ## Default parameters to send to solr for all search-like requests. See also SearchBuilder#processed_parameters
     config.default_solr_params = {
+  :qt => 'search',
```
bibliotechy: I'd like to keep these declarations consistent, so let's use the `qt: 'search',` for instead. Otherwise a developer down the road may be confused as to why this is declared differently.

sensei100:
That doesn't actually need to be there at all, so I removed it.

PR #56, opened 2017-08-01
Update ruby-marc version for bugfix

bibliotechy:
Fixes an issue where traject would bomb out on certain records, leaving any subsequent records 
in the same file unindexed.

Fixes #BL-33


bibliotechy:
The actual change is in https://github.com/ruby-marc/ruby-marc/commit/d0b248e7808c02bc1293467d260a749339cfd5d2

PR #57, opened 2017-08-01
Bl 23 solr field line breaks

sensei100:
Multi-valued fields (including links) are displayed as lists with line breaks.

### DIFF
--- a/spec/fixtures/features.yml
+++ b/spec/fixtures/features.yml
@@ -41,7 +41,7 @@ creator_100:
   creator: "Meḳler, D. L., Subfield B. Subfield C. Subfield D. Subfield E. Subfield J. Subfield L. Subfield M. Subfield N. Subfield O. Subfield P. Subfield Q. Subfield R. Subfield T. Subfield U."
 creator_100_v:
   doc_id: "991012172649703811"
-  creator_vern: "מעקלער, ד.ל. and לאועב, מיטשעל"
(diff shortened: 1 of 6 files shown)


solr/conf/solrconfig.xml
```diff
@@ -54,16 +54,12 @@
...
-  <updateHandler class="solr.DirectUpdateHandler2">
-    <autoCommit>
-      <maxDocs>1000000</maxDocs>
-      <maxTime>900000</maxTime>
-      <openSearcher>true</openSearcher>
-    </autoCommit>
-  </updateHandler>
-
-
+
```
bibliotechy: These changes shouldn't be in here.

spec/fixtures/features.yml
```diff
@@ -41,7 +41,7 @@ creator_100:
   creator: "Meḳler, D. L., Subfield B. Subfield C. Subfield D. Subfield E. Subfield J. Subfield L. Subfield M. Subfield N. Subfield O. Subfield P. Subfield Q. Subfield R. Subfield T. Subfield U."
 creator_100_v:
   doc_id: "991012172649703811"
-  creator_vern: "מעקלער, ד.ל. and לאועב, מיטשעל"
+  creator_vern: "מעקלער, ד.ל.  לאועב, מיטשעל"
```
bibliotechy: Do we want to actually test that these are being displayed in a list? This would pass if they were jut space separated, or new line separated.

PR #59, opened 2017-08-02
Index format facet

skng5:
Issue #BL-5


skng5:
Renaming and addressing BL-5 subtasks.

PR #73, opened 2017-08-21
BL-72 create rake task to harvest incremental oai updates

skng5:
Addresses Issue #BL-72


bibliotechy:
This only harvests the first page of results, right? I don't see any handling of resumption tokens in here. Is it worth dropping in the [ruby-oai client](https://github.com/code4lib/ruby-oai) to deal with that?

skng5:
@bibliotechy Yes, only the first page. The ruby-oai client is well worth pursuing.

skng5:
@bibliotechy - I've added in the resumption token without the OAI gem. It's harvesting based on the resumption token. There are thousands of records, and I'm waiting to get to the end of the list of files so I can verify the loop exit condition.

bibliotechy:
Superseded by #83

PR #74, opened 2017-08-22
Updates to backlog issues

sensei100:
Made some tweaks to Jira issues from the last sprint.


app/assets/stylesheets/tucob.css.erb
```diff
@@ -21,5 +21,11 @@ ul {
 }
 
 .document dt {
-	overflow: visible;
+  white-space: normal;
+  word-break: break-word;
+  width: 170px;
```
bibliotechy: @sensei100  OK, the only issue I have with this is that it doesn't account for smaller screens - we should do some css negotiations based on screen size. This SO example is pretty illustrative - https://stackoverflow.com/a/29388888

sensei100:
Okay, I made the adjustment for smaller screens.

PR #83, opened 2017-09-05
Bl72a create rake task to harvest incremental oai updates

skng5:
Addresses Issues #BL-72 and #BL-73


lib/tasks/oaiharvest.rake
```diff
@@ -0,0 +1,116 @@
...
+
+namespace :fortytu do
+
+  desc 'Posts fixtures to Solr'
+  task :ingest, [:marc_file] => :environment do |t, args|
+    `traject -c app/models/traject_indexer.rb #{args[:marc_file]}`
+    `traject -c app/models/traject_indexer.rb -x commit`
+  end
+
+  desc 'Delete fixtures from Solr'
```
bibliotechy: This seems to duplicate https://github.com/tulibraries/tul_cob/blob/bb5ccea371659d7435d10546f7069bb018afa9fb/lib/tasks/tulsearch.rake#L12-L16 Is there a reason for it?

lib/oai/alma.rb
```diff
@@ -0,0 +1,101 @@
...
+      oai_url << "&from=#{from_time}" unless from_time.empty?
+      oai_url << "&to=#{to_time}" unless to_time.empty?
+
+      harvest_files = []
+
+      begin
+        loop do
+          logger.info("Retrieving #{oai_url}")
+          puts "Retrieving #{oai_url}"
+          client = OAI::Client.new oai_url
```
bibliotechy: This client does not appear to be used?

PR #85, opened 2017-09-08
Update facets

sensei100:
Added the logic into traject for the library  availability facets.

### DIFF
--- a/app/views/catalog/_show_default.html.erb
+++ b/app/views/catalog/_show_default.html.erb
@@ -3,6 +3,7 @@
   <div class="availability">
     <dl class="dl-horizontal  dl-invert">
       <dt class="blacklight-availability show-page-avail">Status/Location:</dt>
+      <span id="alma_availability_url" data-url="<%= alma_availability_url(format: :json) %>" ></span>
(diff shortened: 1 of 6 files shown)


spec/features/indices_spec.rb
```diff
@@ -39,6 +39,7 @@
     context "publicly available pages" do
       scenario "User visits home page" do
         visit '/'
+        save_and_open_page
```
bibliotechy: @sensei100 Is this an artifact of testing?

app/views/catalog/_show_default.html.erb
```diff
@@ -3,6 +3,7 @@
   <div class="availability">
     <dl class="dl-horizontal  dl-invert">
       <dt class="blacklight-availability show-page-avail">Status/Location:</dt>
+      <span id="alma_availability_url" data-url="<%= alma_availability_url(format: :json) %>" ></span>
```
bibliotechy: Is this part of this commit?

bibliotechy: Oh, never mind, I remember now, this is going on the show page

sensei100: Yeah, sorry about that. I probably should have put it in a different branch.

PR #91, opened 2017-09-18
Bl 110 make all fields searchable

skng5:


skng5:
We need to address searching Physical Description, as it's a displayed field.

bibliotechy:
So, the schema *is* searching Physical Description, but it is having trouble parsing the bizarre shorthand `x, 153 p. : ill. ; 25 cm.` otherwise known as MARC.

Seriously, if someone is searching for `illness` do we really want them getting all the records with `ill.` in the Physical Description. Seems like noise to me.

PR #102, opened 2017-10-02
Bug fixes

sensei100:
This addresses the problem with the four digit year method and also adds a line break for field lists.


bibliotechy:
Woohoo better test coverage.

PR #106, opened 2017-10-06
Advanced search config

sensei100:
Addresses several issues relating to advanced search:
- BL-135 Remove facet limits from search results
- BL-138 Edit dropdown for simple search
- BL-131 Edit text in search tips


skng5:
Looks good.

PR #109, opened 2017-10-10
Add docker setup.

dkinzer:
Allows for an alternative set up of the local development environment
using docker.

REF: [BL-139](https://tulibdev.atlassian.net/browse/BL-139)

QA Steps
===
- [ ] install and enable docker
- [ ] clone this repository and cd into it.
- [ ] Run `rake docker:up`


sensei100:
It is not creating a solr core for me.

dkinzer:
@sensei100 I figured out the issue was that I was mounting over the solr cores with a host folder.  It should now work.

dkinzer:
@sensei100 I've made some additional changes to this PR per your suggestions to me this morning.

* I updated the build step so that the fixture data is ingested on initialization.
* I replaced the old alma sample data with a more current dump.
* I updated the README to reflect the new changes.

PR #110, opened 2017-10-10
Footer

cdoyle-temple:
Added Temple style footer to all pages.


app/search_engines/bento_search/primo_engine.rb
```diff
@@ -3,7 +3,7 @@
 class BentoSearch::PrimoEngine
   include BentoSearch::SearchEngine
 
-  @@per_page = 10
+  @@per_page = 15
```
bibliotechy: Why is this changed as part of the footer?

bibliotechy:
@cdoyle-temple Other than the per page thing mentioned above, looks good.

PR #112, opened 2017-10-14
Move script to bin folder.

dkinzer:
By design a Rails app bin folder contains the rails script that starts
the app and can contain other scripts needed to setup, update, deploy or
run the application.

Thus, it does not make sense to have both a scripts folder and a bin
folder. This change moves the lone script in the current application
into the bin folder where other rails developer would look for it by
convention.


bibliotechy:
Yup, makes sense.

PR #119, opened 2017-10-19
Bl 153 use enriched bibs

sensei100:
BL-153 and BL-145 Use enriched bib records rather than the alma api for electronic resources

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -42,7 +42,38 @@ module ApplicationHelper
     new_link.join("<br />").html_safe
   end
 
-  def bento_engine_nice_name(engine_id)
-    I18n.t("bento.#{engine_id}.nice_name")
+  def alma_build_openurl(query)
+    query_defaults = {
+      rfr_id: 'info:sid/primo.exlibrisgroup.com',
+    }
+
+    URI::HTTPS.build(
+        host: alma_domain,
+        path: "/view/uresolver/#{alma_institution_code}/openurl",
+        query: query_defaults.merge(query).to_query).to_s
+    end
+
+    def alma_electronic_resource_direct_link(portfolio_pid)
+      query = {
+          'u.ignore_date_coverage': 'true',
+          'Force_direct': true,
+          portfolio_pid: portfolio_pid
+      }
+      alma_build_openurl(query)
+    end
+
+    def electronic_resource(args)
+      new_link = args[:document][args[:field]].each_with_index.map { |field, i|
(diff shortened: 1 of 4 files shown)


app/helpers/application_helper.rb
```diff
@@ -42,7 +42,38 @@ def electronic_access_links(args)
...
-    I18n.t("bento.#{engine_id}.nice_name")
+  def alma_build_openurl(query)
+    query_defaults = {
+      rfr_id: 'info:sid/primo.exlibrisgroup.com',
+    }
+
+    URI::HTTPS.build(
+        host: alma_domain,
+        path: "/view/uresolver/#{alma_institution_code}/openurl",
+        query: query_defaults.merge(query).to_query).to_s
```
bibliotechy: This looks like it contains a closing paren with not opening paren.

app/helpers/application_helper.rb
```diff
@@ -42,7 +42,38 @@ def electronic_access_links(args)
...
+    def alma_electronic_resource_direct_link(portfolio_pid)
+      query = {
+          'u.ignore_date_coverage': 'true',
+          'Force_direct': true,
+          portfolio_pid: portfolio_pid
+      }
+      alma_build_openurl(query)
+    end
+
+    def electronic_resource(args)
```
bibliotechy: Could this method get a name that gives a little more context about what it does?

Also, it feels like there is a lot going on in this method that maybe could be broken out into a few semantically named methods that might help someone else understand the logic. Like `if args[:value][i].tr('|', '').scan(/\D/).empty?` could be `string_contains_pipe?(args[:value][i])` (which is what I think that is doing), and that might make it a bit more readable.

app/helpers/application_helper.rb
```diff
@@ -42,7 +42,38 @@ def electronic_access_links(args)
...
+      new_link = args[:document][args[:field]].each_with_index.map { |field, i|
+        if args[:value][i].tr('|', '').scan(/\D/).empty?
+          content_tag(:li, link_to("Find it online", alma_electronic_resource_direct_link(args[:value][i].split("|").first)), class: "list_items")
+        else
+          content_tag(:li, link_to(args[:value][i].split("|").second, alma_electronic_resource_direct_link(args[:value][i].split("|").first)), class: "list_items")
+        end
+       }
+      new_link.join("<br />").html_safe
```
bibliotechy: I'd love a test for this helped method please.

app/helpers/application_helper.rb
sensei100: I think it is on line 50

PR #120, opened 2017-10-19
Improve PCI data in Bento Search.

dkinzer:
REF: BL-141

 ## Description:

Update the bento box results to match the fields that appear currently in
Blacklight as closely as possible.

```
From Primo Normalized XML: Main link: display/vertitle, display/title
Second line: display/creator, display/contributor Third line (Label -
"Published"): display/publisher, display/date
```

* Add the ability to configure labels for bento search items.
* Adds publisher to blacklight and primo search result items.
* Adds tests for bento search engines.
* Adds test for the new bento decorator class.

QA Steps
===
- [x] Do a bento search and verify that the following fields are part of
the search result items:

```
:title, :authors, :publisher,
```


.gitignore
```diff
@@ -6,6 +6,7 @@
 
 # Ignore bundler config.
 /.bundle
+*tags
```
bibliotechy: I tend ignore editor artefacts in my global gitignore so every rep doesn't need every devs preferences.

config/bento.yml.example
```diff
@@ -1,8 +1,13 @@
 
 
 default: &default
+  expected_fields:
```
bibliotechy: Why are we adding this to config when it only get used in tests? If you want to use them in multiple tests, maybe the spec_helper would be a batter place?

dkinzer: Is there a better place I can add this info?  It needs to be central so that any future engine tests against the same fields.

.gitignore
dkinzer: go it.

config/bento.yml.example
bibliotechy: `spec_helper`? That gets included in all tests.

PR #123, opened 2017-10-25
Set up Shibboleth Integration.

dkinzer:
REF BL-147

Description
===
We want to handle user authentication via Shibboleth. To that end it would be
nice be able to also test this system locally.

This PR adds a new sp container to the docker setup that runs the app in
development mode behind an apache2 sever that is protected by an SP daemon.

Currently the setup has been configured to work with a test idp server, not
the temple idp server.

QA Steps.
==
* Add `127.0.0.1	tul_cob.temple.edu` to your `/etc/hosts` file
* Start up the containers.
* go to `https://tul_cob.temple.edu` sign in page.
* Follow the instructions to sign in.
- [ ] A new user gets created when you sign in after being redirected by the IdP server.


config/initializers/devise.rb
```diff
@@ -7,6 +7,7 @@
   # Devise will use the `secret_key_base` as its `secret_key`
   # by default. You can change it below and use your own secret key.
   # config.secret_key = '886905c36911440800fb98070d37b6ae1b5a7dc535bbdaefe15fe1ce0158fd5e2f5bf4590e983f4d9d741d7a06b640e2987325fd23fb4ee8a2d2a8aadce69787'
+  config.secret_key = '664b16e18fa6e3d8beda1f2af71b836f081c5dd2c1083ab0febd0a3f5e49c1fb32eedc1518f2c95cb84b99a8f8716503e3a4e2d8a382b6dca497248fd2137240'
```
bibliotechy: @dkinzer I don't think we want to do this in production

dkinzer: I'll remove that... I think I added it when I didn't realized that the site was running as production by default.

PR #126, opened 2017-10-27
Add tests to an overriding function.

dkinzer:
This commit adds some tests to a custom override of
`alma_service_type_for_fulfillment_url`

### DIFF
--- a/app/helpers/blacklight_alma_helper.rb
+++ b/app/helpers/blacklight_alma_helper.rb
@@ -1,13 +1,17 @@
-module BlacklightAlmaHelper
+# frozen_string_literal: true
 
+module BlacklightAlmaHelper
   include BlacklightAlma::HelperBehavior
 
-  def alma_service_type_for_fulfillment_url(document)
-    return false if document['availability_facet'].nil?
-    if document['availability_facet'].include?("Online")
-      'viewit'
+  # Overrides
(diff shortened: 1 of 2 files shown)


app/helpers/blacklight_alma_helper.rb
```diff
@@ -1,13 +1,17 @@
...
-  def alma_service_type_for_fulfillment_url(document)
-    return false if document['availability_facet'].nil?
-    if document['availability_facet'].include?("Online")
-      'viewit'
+  # Overrides
+  # BlacklightAlma::HelperBehavior#alma_service_type_for_fulfillment_url.  Bib
+  # records may have both physical and electronic holdings: This guarantees
+  # that we always "viewit" in the case of an "Online" record.
+  def alma_service_type_for_fulfillment_url(document = {})
+    if document.to_h["availability_facet"].to_a.include?("Online")
```
bibliotechy: @dkinzer Am I correct that the `document.to_h` is there to ensure you get a hash even if document is `nil` to avoid an Exception? I feel like that intention obscured a bit. I feel like it might be a little easier to understand with something like:
```ruby
def alma_service_type_for_fulfillment_url(document)
  document ||= {}
  document.fetch("availability_facet", []).include?("Online")
...
```

dkinzer: This bit can still return nil `document.fetch("availability_facet", [])` so it does require to_a there too.

PR #129, opened 2017-10-30
BL-162 Links from 856 field and alma api are merged into one list

sensei100:
BL-162 Links coming from the alma api and links coming from Marc 856 fields are merged into one field.

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -167,7 +167,8 @@ class CatalogController < ApplicationController
     # solr fields to be displayed in the show (single result) view
     #   The ordering of the field names is the order of the display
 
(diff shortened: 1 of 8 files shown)


bibliotechy:
@jennifer
I downloaded a random sample of records and I got an error when trying to do do a search
screenshot-2017-11-1 action controller exception caught
The document object did not have an `availability_facet` key.

sensei100:
@bibliotechy  How do we want to handle records that might not have the availability_facet key?

bibliotechy:
I'm confused why there would be records without some value. Seems like every record should either be Online or At the Library.

bibliotechy:
I'm seeing multiple Online buttons for records with more than one external link. Pretty sure there should only be one. 
screenshot-2017-11-1 availability online - blacklight search results

sensei100:
I don't see any multiple buttons on mine, so I'm not sure what is causing that.

PR #130, opened 2017-10-30
BL-172: Make traject custom code more testable.

dkinzer:
REF BL-172

Description
===
The current code that is used to configure the Traject indexing is mostly stuck
inside of blocks that are growing in size and cluttering the
configuration file.

By moving these blocks into higher order methods, we accomplish
several beneficial changes:
* The configuration will be easier to read and follow.
* The code that is currently in block format will be more easily
testable.

QA Steps
===
- [ ] Run a traject import and verify that the import is still viable:
`docker-compose exec app traject -c app/models/traject_indexer.rb
sample_data/alma_bibs.xml`


dkinzer:
haha... github wants me to redo the whole thing.

bibliotechy:
Mostly looks good.

My only concern is that `lib/traject/macros/marc_format_classifier.rb` is specifically an override of [the traject marc_format_classifier](https://github.com/traject/traject/blob/master/lib/traject/macros/marc_format_classifier.rb). So adding in all of our custom code there feels wrong.

Maybe add a `lib/traject/macros/tul_utils.rb` or something like that to break it out so it is clear what we are overriding and what is just our own stuff?

Whaddya think?

dkinzer:
That was one of my issues I was concerned about too.  And originally I created another file and was including it in the model file.  But for some reason some of the bits that it requires weren't getting pulled into the completely custom file.  I decided to just go with it for now instead of trying to fight it.

I guess I could try to figure out how I an merge any of the required dependencies.

dkinzer:
I wish also that the autoload for this stuff would make more sense to me.  Not sure why we need to require files all over the place and not just be able to include our modules and classes and have rails know where to look for them.  Is there a guide somewhere of how to set that up properly?

bibliotechy:
So, part of the issue is that, by design, the `traject` command doesn't need to know about rails. We bundle it up with the rails app for simplicity, but it could be a standalone config file run on a machine that is not running rails. Because of tat, it doesn't get any of the rails autoloading magic. It's the proce we pay for loosely coupling it.

bibliotechy:
Traject has [some docs](https://github.com/traject/traject/blob/master/doc/indexing_rules.md#macros) on creating your own macros. Have you seen that? Might that help in creating the standalone?

dkinzer:
I have seen that, it's not useful in my case since we are already doing all that.  There is not guide there as to how to include the macro properly. But I have an idea now (maybe my first attempt I was trying to rely on some autoload magic... I can't remember now).

PR #132, opened 2017-11-01
Add Simple rake ingest task

bibliotechy:
Add a rake task to ingest a marc file with traject, and send the commit request on completion.


dkinzer:
Beautiful!  Do you think we should pass a default path to the xml since it's almost always going to be the sam thing and only override it if the rake command is invoked with arguments?

bibliotechy:
Sounds reasonable.

PR #133, opened 2017-11-01
BL-160 Update button to display Loading until alma api data is ready

sensei100:
BL-160 Update availability buttons for physical holdings
- Button now displays Loading until it receives data from the alma api
- Button changes the class and text for each physical holding based on its availability data


bibliotechy:
Does this need to be paired with a change to `alma_blacklight`?

sensei100:
Yes.

PR #134, opened 2017-11-02
Move search params defaults to catalog_controller

bibliotechy:
REF: BL-124

Description
===
Move the solr query parameter configs like ps, pf, author_qf, etc, out of the solrconfig searchHandler and into the catalogContoller
Temporarily move this configuration into the catalog controller for ease of use and update. Once our work on relvancy and schema settles down, we can move it back into solrconfig.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,6 @@ git clone git@github.com:tulibraries/tul_cob
 cd tul_cob
 bundle install
 cp config/secrets.yml.example config/secrets.yml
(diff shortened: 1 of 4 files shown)


dkinzer:
@bibliotechy from the description it sounds like there should be an equivalent number of removal of code to addition of code but I see more code added than deleted.  Is this PR missing something or is that just additional configuration?

PR #137, opened 2017-11-08
Only add separator if has date statement in e resource link

bibliotechy:
The separator between e resource links and their data statements was appearing even if the date statement was missing.  Something like:
screenshot-2017-11-8 availability online - blacklight search results

This pull request makes that go away!!!

### To Verify
* Fetch and checkout his branch
* Run your fave variation on `bundle exec rake server`
* Go to the show page for record with id `991013867929703811`

- [ ] Inside the availability section at the bottom of the page, there should be a link to EbscoHost Academic Ebook Collection that does not have a trailing  "-" (dash).

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -65,6 +65,10 @@ module ApplicationHelper
       query: query_defaults.merge(query).to_query).to_s
   end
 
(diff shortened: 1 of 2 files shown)


sensei100:
Looks good.

PR #138, opened 2017-11-09
Online and physical holdings

sensei100:
BL-169 Adjust availability settings
- Refactor single_link_builder method to work with both alma api links and 856 fields
- Refactor alma_service_type_for_fulfillment method to display physical holdings first
- There is a change in the blacklight_alma gem that also goes along with this


bibliotechy:
Related to https://github.com/tulibraries/blacklight_alma/pull/8

PR #140, opened 2017-11-13
Aeon integration

sensei100:
BL-174 Aeon Integration with two subtasks
- BL-181 Builds a link to the aeon request page and pre-populates the form with data.  You will need to sign up for a SCRC Researcher Account Logon to test that this works.
-BL-182 A button to make an aeon request only shows on records that are in the SCRC and are located in rare stacks.
screen shot 2017-11-13 at 11 03 13 am
- http://localhost:3000/catalog/991019807439703811 should display the link
- http://localhost:3000/catalog/991030814199703811 should not have the link


app/helpers/application_helper.rb
```diff
@@ -104,4 +105,25 @@ def single_link_builder(field)
   def bento_engine_nice_name(engine_id)
     I18n.t("bento.#{engine_id}.nice_name")
   end
+
+  def aeon_request_url(document)
+    openurl_field_values = {
+      ItemTitle: document["title_statement_display"].first.to_s,
```
dkinzer: There is some repetition here that you could avoid by first codifying the key changes, then reducing that value to get the new one .  `[ItemPlace: "imprint_display", ReferenceNumber: "alma_mms_display", ..].inject {|acc, key| ...}`

dkinzer:
I might be missing something, but I could not reproduce this locally.

sensei100:
Could you be more specific about what is not working for you?

dkinzer:
The link does not show up for me when I go to `catalog/991019807439703811`

sensei100:
Did you re-ingest your records?  There were changes made to traject that need to be pulled in.

bibliotechy:
OK, the issue was that this PR included a change to the Solr schema. Using solr wrapper locally, the change was not picked up, because solr wrapper copies `solr/conf` to a temp directory where solr lives when it is first set up. Unless we call solr_Wrapper clean, it is not pulled in. Not sure how your docker stuff handles that, @dkinzer, but one I started with a fresh solr, the button appeared.

bibliotechy:
Also, the button does what it purported to do, fills in the form with details.

Looks good!

dkinzer:
In the docker instance I mount solr/conf over the docker instance's solr configuration folder, so that should not be an issue.

PR #141, opened 2017-11-13
BL-180 Add new skin for alma iframe

sensei100:
BL-180 Allows us to create our own custom css for the alma availability iframe.  The delivery_domain in the alma.yml.example file was also out of date, so that updated to use the sandbox.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -56,8 +56,6 @@ module ApplicationHelper
 
   def alma_build_openurl(query)
     query_defaults = {
-      'is_new_ui': true,
-      "req.skin": "temple_01",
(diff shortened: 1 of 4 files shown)


dkinzer:
This looks good, though I can't figure out where it's getting used so I can't verify that it works.

dkinzer:
When I look at this locally I can see that the hostname has changed, but I cannot see that req.skin is being added as a url-parameter:

```<iframe class="bl_alma_iframe" src="https://sandbox01-na.alma.exlibrisgroup.com/view/uresolver/01TULI_INST/openurl?rfr_id=info%3Asid%2Fprimo.exlibrisgroup.com&amp;rft.mms_id=991030803889703811&amp;svc_dat=getit"></iframe>```

PR #142, opened 2017-11-14
BL-176: Add staff view link to document view.

dkinzer:
REF BL-176

Description
==
Add `Staff View` link to document view that takes user to Raw Marc Record view for the document.

QA Steps
===
* Do a search and click on a document.
- [ ] Verify that there is a `Staff View` link at the bottom of the page.
* Follow the link.
- [ ] Verify that page includes a copy of pretty printed XML and the the title is "Staff View."


app/controllers/catalog_controller.rb
```diff
@@ -9,8 +9,16 @@ class CatalogController < ApplicationController
...
   include BlacklightAlma::Availability
 
+  include Blacklight::Marc::Catalog
+
   #helper BlacklightAlma::HelperBehavior
 
+  # Overrides CatalogController::show in order to add extra instance variables.
+  def show
+    super
+    @staff_view_title = t("blacklight.search.librarian_view.title")
```
bibliotechy: Why not just call the `t` method directly in the view?

dkinzer: Because I want to keep the view as clean as possible.

dkinzer: @bibliotechy if I were to take care of that in the view, then I would have to move the logic that checks whether or not there is full marc available to the view.  This way I can leave the logic outside of the view and just let the view take care of what it knows.

bibliotechy: But, the [rails guide on i18n](http://guides.rubyonrails.org/i18n.html) gives examples with the `t` method in the template. It's standard rails convention that is used in lots of places.

dkinzer: Oh, actually I was referring to something else.  You're absolutely right.

dkinzer: I think my main motivation for doing this was to make it easy for me to test that I was using the correct link name

dkinzer: But I realize now I can just test that t("blacklight.search.librarian_view.title") is set to what I want.

bibliotechy:
So, in order to make the view as simple as possible, we diverge from the blacklight-marc code base and are now on the hook maintaining changes to the overridden code's API.
 
It feels like we are trading imperfect view implementation for technical debt. I'm not convinced it is worth the trade off.

dkinzer:
@bibliotechy actually we can't use the code in blacklight-marc at all.  Instead of overriding I should really be creating a new action.  The reason we can't use it is that our solr docs do not respond to a method that is required by the blacklight-marc view and helper functions.

bibliotechy:
Which method?

dkinzer:
@bibliotechy https://github.com/projectblacklight/blacklight-marc/blob/v6.2.0/app/views/catalog/librarian_view.html.erb#L6 `:to_marc`.

dkinzer:
When I was trying to use their code I would just get a blank on that page.

PR #147, opened 2017-11-16
Add masquerade functionality

skng5:
Issue #BL-178

PR #150, opened 2017-11-17
Add masquerade functionality (Attempt 3)

skng5:
Rebase issues with PR #147 and #148
Resubmission for issue #BL-178

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -1,6 +1,22 @@
 # frozen_string_literal: true
 
 class UsersController < ApplicationController
+
+  def index
+    @users = User.order(:id)
+  end
+
+  def impersonate
+    user = User.find(params[:id])
+    impersonate_user(user)
+    redirect_to root_path
+  end
+
+  def stop_impersonating
+    stop_impersonating_user
+    redirect_to root_path
+  end
+
   def account
     @loans = current_user.get_loans
     @holds = current_user.get_holds
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -2,9 +2,14 @@
 
 class UsersController < ApplicationController
   before_action :require_admin!, only: [:index]
+  before_action :require_non_production!, only: [:index]
 
   def require_admin!
     redirect_to root_path unless current_user && current_user.admin
+  end
+
+  def require_non_production!
+    redirect_to root_path if Rails.env.production? #&& !Rails.config.allow_impersonator
   end
 
   def index
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -9,7 +9,7 @@ class UsersController < ApplicationController
   end
 
   def require_non_production!
-    redirect_to root_path if Rails.env.production? #&& !Rails.config.allow_impersonator
+    redirect_to root_path if Rails.env.production? && !Rails.configuration.allow_impersonator
   end
 
   def index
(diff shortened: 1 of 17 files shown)


dkinzer:
@skng5 that error looks like it's an issue with a javascript error?  Maybe is a javascript dependency on travis?

skng5:
Problem only occurs when `Rails.env = development`. It may be that this is not a reasonable test for a situation that runs only in development mode.

skng5:
Removed test in dev environment. Passes without out error. _Good to go._

..finally...

app/views/layouts/blacklight.html.erb
```diff
@@ -30,6 +30,12 @@
 
   <%= render partial: 'shared/ajax_modal' %>
 
+  <% if false %>
```
bibliotechy: @skng5 I'm not sure why this is here. Won't this never happen?

config/application.rb
```diff
@@ -20,6 +20,7 @@ class Application < Rails::Application
     config.alma = config_for(:alma).with_indifferent_access
     config.bento = config_for(:bento).with_indifferent_access
     config.exceptions_app = routes
+    config.allow_impersonator = false
```
bibliotechy: We need a way to override this in deployments. Maybe add an `ENV["allow_impersonator"] ||= false` so we can override with a env veriable in an ansible delpoyment?

spec/controllers/users_controller_spec.rb
```diff
@@ -80,4 +80,93 @@
...
+      context "Configured impersonator allowed" do
+        it "shows users page" do
+          expect(subject).to render_template("users/index")
+        end
+      end
+    end
+
+    context "Production environment" do
+      before :each do
+        Rails.env = "production"
```
bibliotechy: I thought you removed the tests that set the environment. Just double checking that this should be here.

app/controllers/users_controller.rb
```diff
@@ -1,6 +1,32 @@
...
+  def require_admin!
+    redirect_to root_path unless current_user && current_user.admin
+  end
+
+  def require_non_production!
+    redirect_to root_path if Rails.env.production? && !Rails.configuration.allow_impersonator
+  end
+
+  def index
+    @users = User.order(:id)
```
bibliotechy: So, this will return all of the anonymous users Blacklight stores for recording search sessions, who we are not interested in masquerading as.  
We could only select users that are not guests? `User.where.not(guest: true)`

skng5: Yes, it's the whole kit and caboodle. We can try returning non-guests though.

app/views/layouts/blacklight.html.erb
skng5: That's development debris.  Ooops!

config/application.rb
skng5: Good idea.

spec/controllers/users_controller_spec.rb
skng5: I removed the Development environment test. Production is handled differently than test and the formerly tested development environment.

app/controllers/users_controller.rb
bibliotechy: That sounds like a good idea.

spec/controllers/users_controller_spec.rb
bibliotechy: 👍

skng5:
@bibliotechy Ready for review.

spec/controllers/users_controller_spec.rb
skng5: Ready for review

PR #151, opened 2017-11-17
Add interpolation to .nil?

sensei100:
Running bundle exec rake rspec was throwing an error.  This change allows the tests to run.

### DIFF
--- a/lib/tasks/server.rake
+++ b/lib/tasks/server.rake
@@ -19,7 +19,7 @@ task :rspec, [:spec_args] do |t, args|
   if Rails.env.test?
     run_solr("test", port: "8985") do
       Rake::Task["fortytu:solr:load_fixtures"].invoke
-      rspec_cmd = "rspec" + (" #{args[:spec_args]}" unless args[:spec_args].nil?)
+      rspec_cmd = "rspec" + (" #{args[:spec_args]}" unless "#{args[:spec_args]}".nil?)
       system(rspec_cmd)
     end
   else
--- a/lib/tasks/server.rake
+++ b/lib/tasks/server.rake
@@ -19,7 +19,7 @@ task :rspec, [:spec_args] do |t, args|
   if Rails.env.test?
     run_solr("test", port: "8985") do
       Rake::Task["fortytu:solr:load_fixtures"].invoke
-      rspec_cmd = "rspec" + (" #{args[:spec_args]}" unless "#{args[:spec_args]}".nil?)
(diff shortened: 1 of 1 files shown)


lib/tasks/server.rake
```diff
@@ -19,7 +19,7 @@ task :rspec, [:spec_args] do |t, args|
   if Rails.env.test?
     run_solr("test", port: "8985") do
       Rake::Task["fortytu:solr:load_fixtures"].invoke
-      rspec_cmd = "rspec" + (" #{args[:spec_args]}" unless args[:spec_args].nil?)
+      rspec_cmd = "rspec" + (" #{args[:spec_args]}" unless "#{args[:spec_args]}".nil?)
```
dkinzer: I don't think this will ever be true
```ruby
irb(main):001:0> "#{nil}".nil?
=> false
irb(main):002:0> ```

**Comment by sensei100** [comment]
I'm sorry I was misleading in my commit message.  Last week we had a frozen string literal error that was blocking tests from running.  We removed a shovel operator and added  + (" #{args[:spec_args]}" This got the tests running when an argument was passed in, but was failing without.  This commit adds the interpolation to the second part of the statement unless args[:spec_args].nil?) to allow tests to pass in both situations.

**Comment by bibliotechy** [comment]
@sensei100 I think @dkinzer's point is that when you interpolate anything with `#{}` it becomes a string, so it will never evaluate to `nil`, so `nil?` will always return false. 

So, with this change, when no `:spec_arg` is passed, 
```ruby 
(" #{args[:spec_args]}" unless "#{args[:spec_args]}".nil?)
```  
will return `" "`, which works, cause it doesn't break the task from running, but it does kind of obfuscate the logic. 

The reason it was failing in the first place was because when you don't pass a spec_arg to 
```ruby
(" #{args[:spec_args]}" unless args[:spec_args].nil?)
``` 
that parenthesis returns `nil` and when you try to add nil to a string, you get `TypeError: no implicit conversion of nil into String`. Pretty sure I suggested that parenthesis when we paired on debugging that, so, 🤚 <-- that one's my mistake.

It is a really common pattern, of only adding to a string when some value is not `false` or `nil`, and I think making explicit what is happening might be a easier approach to understand in the future when some other bug inevitably crops up. Something like 

```ruby
["rspec" , args[:spec_args]].compact.join(" ")
```
would correctly build the string whether or not `:spec_args` is present, and not add any extra spaces when they are not needed.

sensei100:
@bibliotechy This is now revised according to your suggestion.

PR #155, opened 2017-11-29
Link to lib account only shows when logged in.

bibliotechy:
Fixes path to account when logged in.


dkinzer:
LGTM (was not able to test locally because my created user wouldn't log in for some reason, but I don't think that's a real issue since the tests look like they cover logging in).

PR #156, opened 2017-11-30
Bl 188 update library facet

sensei100:
BL-188 Updates to library facet
- Changed the mappings in the locations_map according to the documentation
- Interlibrary Loan should not show up in the list of facet values


lib/traject/macros/custom.rb
```diff
@@ -168,6 +168,15 @@ def normalize_lccn
         end
       end
 
+      def extract_library
+        lambda do |rec, acc|
+          rec.fields(["HLD"]).each do |field|
+            subfield_b = field["b"] if field["b"] != "RES_SHARE"
+            acc << Traject::TranslationMap.new("locations_map")[subfield_b]
```
bibliotechy: If there is only one item, and it has `"RES_SHARE"` in `field["b"]`, then `subfield_b` won't be defined when you call `Traject::TranslationMap.new("locations_map")[subfield_b]` and it will throw an error, or return something unexpected, like the entire `locations_map` Hash.

sensei100:
@bibliotechy I adjusted the logic according to your recommendation.

PR #166, opened 2017-12-05
Use relative path for Back to Main View link.

dkinzer:
REF BL-176

The "Back to Main View" link does not work in
http://libqa.library.temple.edu/catalog/catalog/991024675119703811/staff_view
where the path to the catalog is double labeled.

This commit fixes this issue by making the path back to the main view
relative to the current path.

QA Steps
==
* Do a catalog search
* Follow a result link to main view
* Follow the staff view link
- [ ] Verify that the url for "Back to Main View" is a relative format link, `../:id`
- [ ] Verify that clicking on the link takes you back to the document main view.


dkinzer:
@skng5 thanks, and good question; I asked the same question on BL-176

PR #171, opened 2017-12-12
Bl 207 solr display fields

sensei100:
BL-207 Update mappings for title_series_display, date_copyright_display, and note_display fields
- title_series: Add several new fields and create specs for them
- date_copyright: Update logic in the extract_copyright macro to include records with a "c", "p", or copyright symbol in 260c and create tests for each case.
- note_display:  Add the 525a field and create test.


dkinzer:
Looks good ... I noticed that regular searches from within admin page are throwing a javascript error but I don't think it has anything to do with this issue since it's also happening on master.

PR #173, opened 2017-12-13
Add a precision operator to advanced searches.

dkinzer:
REF BL-133

Adds a precision operator to allow searches on :is, :begins_with and the
default :contains.

This change requires a Solr reindex to take effect.

Currently *_t fields are automatically flanked with `matchbeginswith`
and `matchendswith`.  On a :begins_with the solr query is changed under
the hood to search for "matchwithbegins foo bar".

The :is "for exact" phrase works in a similar way, by updating the
search value to be quoted instead of free.

## QA Steps
*  Download the PR locally and reindex.
- [ ] Do a an advanced search with :is and :contains and :begins_with and
verity that they work as you may expect them to.

## Additional changes.
* I added a new container called "cli" to docker-compose, this adds an
editor which to containers in cases where we need to attach to the
container and edit something directly we don't have to always download
an editor first.

* I made it so that both cli and sh containers now use the base app
container in order to reuse that container image.

### DIFF
--- a/app/models/search_builder.rb
+++ b/app/models/search_builder.rb
@@ -3,14 +3,115 @@
 class SearchBuilder < Blacklight::SearchBuilder
   include Blacklight::Solr::SearchBuilderBehavior
   include BlacklightAdvancedSearch::AdvancedSearchBuilder
-  self.default_processor_chain += %i[add_advanced_parse_q_to_solr add_advanced_search_to_solr]
   include BlacklightRangeLimit::RangeLimitBuilder
 
-  ##
-  # @example Adding a new step to the processor chain
-  #   self.default_processor_chain += [:add_custom_data_to_query]
-  #
-  #   def add_custom_data_to_query(solr_parameters)
-  #     solr_parameters[:custom] = blacklight_params[:user_value]
-  #   end
+  BEGINS_WITH_TAG = "matchbeginswith"
+  ENDS_WITH_TAG = "matchendswith"
+
+  self.default_processor_chain +=
+    %i[add_advanced_parse_q_to_solr add_advanced_search_to_solr ] +
+    # Process order matters; :begins_with_search assumes :exact_phrase_search
+    # happens after it, see the note for :to_phrase method for extra context.
+    [ :begins_with_search ] +
+    [ :exact_phrase_search ]
+
+
+  def begins_with_search(solr_parameters)
+    dereference_with(:append_start_flank, solr_parameters)
+  end
+
+  def exact_phrase_search(solr_parameters)
+    dereference_with(:to_phrase, solr_parameters)
+  end
+
+  private
+
+    def dereference_with(method, solr_parameters)
+      query = solr_parameters["q"] || ""
+      params = get_params
+
+      # Search misbehaves if we alter non advanced search query.
+      if !query.empty? && !params.empty? && params["search_field"] == "advanced"
+
+        # We need the original values in the search for use in creating
+        # a de-referenced version of the query.
+        fields = get_params.select { |k| k.match(/^q_/) }
+
+        # We de-reference values in order to be able to quote them; otherwise,
+        # solr throws a 500 error: https://stackoverflow.com/a/10183238/256854
+        #
+        # First we generate something like:
+        # [[["{}", "foo", "AND"], "q_1"], [["{}", "bar", "OR"], "q_2"]]
+        # And then we map that to ["_query_:..", "_query_:..."]
+        # @see :parse_queries, and :param_dereference methods below for more
+        # details.
+        queries = parse_queries(solr_parameters["q"])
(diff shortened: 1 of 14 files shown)


bibliotechy:
I was not able to reproduce the expected behavior.

My steps:
* checked out the branch
* solr_wrapper clean
* rake server (ingests only the test fixture data)
* Navigate to advanced search
* select `Title` for field, select `begins with` for precision and enter the search term "Education"

Expected results - only records whose title's start with Education will be in search results

Actual results - Some items start with education, some do not
screen shot 2017-12-15 at 2 57 41 pm

dkinzer:
hmm, i've updated it twice do merge conflicts... maybe that caused an issue.  i'm checking locally.

dkinzer:
OK, i'm seeing the same behavior.  I think that just having the *_t attribute be flank is not enough because the other fields are competing.  I'll need to either boost the relevance of *_t or remove other fields altogether from the search.

dkinzer:
OK, I found a bug I  introduced when I did some cleanup and I had not written a test case for it.  I will push up the test first.

dkinzer:
OK, this should work as expected now.  I should have retested after refactoring it.

dkinzer:
image

dkinzer:
I'm just noticed one other issue.  i'm putting this back on wip

dkinzer:
OK, found the last bug, and added another test case.  Issue was that I was assuming mutations instead of a non mutable state for user input, so my counter mutation logic was messing stuff up.  I added some notes to make some decisions clearer in this regard.

bibliotechy:
Hmmm. Begins with works as expected now for a single field. However, when I added another level on to that I get no results. For example:

image


returns no results.  

I expected it to return:

image


The URL also looks like it is repeating fields.

`http://localhost:3000/catalog?utf8=%E2%9C%93&f_1=title&f_2=all_fields&f_3=all_fields&op_1=AND&op_2=AND&op_row%5B%5D=begins_with&op_row%5B%5D=contains&op_row%5B%5D=contains&q_2=&q_3=&range%5Bpub_date_sort%5D%5Bbegin%5D=&range%5Bpub_date_sort%5D%5Bend%5D=&search_field=advanced&sort=score+desc%2C+pub_date_sort+desc%2C+title_sort+asc&f_1=title&op_row%5B%5D=begins_with&q_1=The&op_1=AND&f_2=creator_t&op_row%5B%5D=contains&q_2=James&op_2=AND&f_3=all_fields&op_row%5B%5D=contains&q_3=&range%5Bpub_date_sort%5D%5Bbegin%5D=&range%5Bpub_date_sort%5D%5Bend%5D=&sort=score+desc%2C+pub_date_sort+desc%2C+title_sort+asc&search_field=advanced&commit=Search`

dkinzer:
The issue with the search you did is the use of  "The" as a search word.  I did not do anything special to the index of `*_t` words other than add the flanks.  When I do a search using title begins with "Education" and author contains "James", I get a single expected result: 

image

Also the code does not do anything to the URL parameters that is all blacklight or blacklight_advanced_search ... (not sure who the culprit is) but if you do a "start over" search you will see the extra fields go away.

bibliotechy:
Sorry, pebcak. I see it working now as expected. :)

app/models/search_builder.rb
```diff
@@ -3,14 +3,98 @@
...
+    end
+
+    def parse_queries(query_string)
+      query_string.split("_query_:")
+        .select { |q| q[0..1] == "\"{" }
+        .map { |q| q.scan(/{(.*)}(.*)\".?(AND|OR|NOT)?/) }
+        .map { |q| q.flatten }
+    end
+
+    def param_dereference(q, k)
```
bibliotechy: I'd love some comments here explaining what q is. What kind of thing is it?

app/models/search_builder.rb
```diff
@@ -3,14 +3,98 @@
...
+
+      # Search misbehaves if we alter non advanced search query.
+      if !query.empty? && !params.empty? && params["search_field"] == "advanced"
+        ops = params.fetch("op_row", [])
+
+        fields = params.select { |k| k.match(/^q_/) }
+
+        # We de-reference values in order to be able to quote them; otherwise,
+        # solr throws a 500 error.
+        queries = parse_queries(query)
```
bibliotechy: I'd love some comments explaining what is happening here. I'm finding it hard to follow the `parse -> zip -> map -> param_dereference` and a little context might help future us grok it.

dkinzer:
I've resubmitted a version with more detailed commenting, and I've added some links to the Jira ticket to the sources I used to create this feature.

bibliotechy:
Sweeeet! Exactly what I was hoping for there!

dkinzer:
Yeah, it was a good suggestion to add more detailed comments there.

PR #174, opened 2017-12-14
Bl 208 update resource type mappings

skng5:
This has the updates to the resource mapping.
Made changes to marc_format_classifier to make operation _clear_ and so it corresponds more closely to the [data mappings documentation](https://tulibdev.atlassian.net/wiki/spaces/SAD/pages/22839300/Data+Mappings+Displays+Facets+Search#DataMappings(Displays,Facets,Search)-ResourceTypeMappings)

The find_xml.rb and get_record.rb are work in progress used to find and create fixtures for testing. This code is low priority, non-production, work-in-progress and doesn't require extensive review.

I would like to have more than one person review this PR.


ebtoner:
I think that this file can be generally understood by non-programmers. A couple of things that I noticed, though: 

Re: marc_genre_007 -- I'm not sure whether we need this mapping or not. Checking with Leanne now and should know by Monday hopefully. 

Re: marc_genre_leader -- 'm' should be computer_file, not database
https://github.com/tulibraries/tul_cob/blob/34b31aef92b47d85f8ff3fc883f3deb063d999b4/lib/translation_maps/marc_genre_leader.yaml#L25

Re: Archival Materials -- please see the mappings that have been removed for this (on Confluence they have been struck out and are in red). I believe that Lines 138 through 143 can be removed: https://github.com/tulibraries/tul_cob/blob/34b31aef92b47d85f8ff3fc883f3deb063d999b4/lib/traject/macros/marc_format_classifier.rb#L138-L143. 

Re: Conference proceedings and Gov Docs -- it looks like these updates haven't been added yet. Is that correct?

ebtoner:
@skng5 Leanne has confirmed that we can remove 007 field mappings for resource types

skng5:
@bibliotechy - I've made all the changes except for Conference Proceedings and GovDocs. @ebtoner will address that after we return from break. It's ready for final review and merge

lib/traject/macros/marc_format_classifier.rb
```diff
@@ -80,49 +73,92 @@ def formats(options = {})
...
 
-        [results].flatten
+      def controlfield_value(controlfield, position, translation_map)
+          controlfield.collect { |f| translation_map[f.value.slice(position)] }
       end
 
       # Just checks if it has a 502, if it does it"s considered a thesis
       def thesis?
         @thesis_q ||= begin
-                        ! record.find { |a| a.tag == "502" }.nil?
```
dkinzer: I'm curious about the reason for changing access syntax to `record` to use the instance form (`@record`) vs attribute reader form `record`.  Shouldn't `record` still work since there is an `attr_reader` set for `:record`.  Is that a readability issue, or is something else going on?

nomadicoder: We missed that between the review session and now. I'll address it now.

skng5:
@dkinzer Done

bin/find_xml.rb
```diff
@@ -0,0 +1,92 @@
...
+  record_ids.each do |id|
+    puts "URL: http://libqa.library.temple.edu/catalog/catalog/#{id}"
+    puts "Command: bin/get_record.rb #{file_path} #{id}"
+  end
+end
+
+def find_by_type(doc, search_type)
+  
+  marc_genre_leader   = Traject::TranslationMap.new("marc_genre_leader").to_hash
+  marc_genre_007      = Traject::TranslationMap.new("marc_genre_007").to_hash
```
dkinzer: assigned but not used.

bin/find_xml.rb
```diff
@@ -0,0 +1,92 @@
...
+  marc_genre_008_21   = Traject::TranslationMap.new("marc_genre_008_21").to_hash
+  marc_genre_008_26   = Traject::TranslationMap.new("marc_genre_008_26").to_hash
+  marc_genre_008_33   = Traject::TranslationMap.new("marc_genre_008_33").to_hash
+  resource_type_codes = Traject::TranslationMap.new("resource_type_codes").to_hash
```
dkinzer: `resource_type_codes`: assigned but not used.

bin/find_xml.rb
```diff
@@ -0,0 +1,92 @@
...
+  doc.remove_namespaces!
+  doc_ids = []
+  doc.xpath('//record//metadata//record').each do |record|
+  #doc.xpath('//record').each do |record|
+    leader = record.xpath('leader').first.text
+    cf008 = record.xpath("controlfield[@tag='008']").text
+    cf006 = record.xpath("controlfield[@tag='006']").text
+    id = record.xpath("controlfield[@tag='001']").text
+    
+    newxml = Nokogiri::XML(record.to_xml)
```
dkinzer: `newxml` assigned but not used (in this function).

bin/find_xml.rb
skng5: Per my original comment in this PR, Not addressing at this time.

bin/find_xml.rb
skng5: Per my original comment in this PR, Not addressing at this time.

PR #175, opened 2017-12-15
Removed redundant fields and added new title fields for search

ebtoner:


bibliotechy:
@sensei100 Can you take a look at [the conflict](https://github.com/tulibraries/tul_cob/pull/175/conflicts) in the traject file and update? I think it was changes that you made in another branch that are now in conflict, so I think your best placed to sort it out.

sensei100:
I fixed the conflict.

PR #178, opened 2017-12-15
BL-206 Move pub date to facet partial and display on one line

sensei100:
BL-206 Update pub date in advanced search
- move publication year code to the facet partial
- display on one line
- update specs to include the new advanced facets


dkinzer:
Perfect, this lgtm, plus is fixes pub date facet not automatically opening when doing an advanced date search!

PR #179, opened 2017-12-20
Fix creator contributor

sensei100:
BL-45 and BL-230 Fixes behavior related to splitting the creator and contributor fields
- Move the creator call inside block so that it displays all creators
- Use trim_punctuation macro for creator_facet
- Use the same regex logic in the browser_creator and creator_index helper methods to display creators and have the links work to creator_facet
- Update fixture file for punctuation

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -30,11 +30,12 @@ module ApplicationHelper
   def browse_creator(args)
     creator = args[:document][args[:field]]
     creator.map do |name|
-      linked_subfield = name.split("|").first.sub(/ *[ ,.\/;:] *\Z/, "")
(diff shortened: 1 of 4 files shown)


app/helpers/application_helper.rb
```diff
@@ -30,22 +30,23 @@ def list_with_links(args)
   def browse_creator(args)
     creator = args[:document][args[:field]]
     creator.map do |name|
-      linked_subfield = name.split("|").first
-      newname = link_to(linked_subfield, root_url + "/?f[creator_facet][]=#{linked_subfield}").html_safe
+      linked_subfields = name.split("|").first.sub(/ *[ ,\/;:] *\Z/, '').sub(/( *[[:word:]]{3,})\. *\Z/, '\1')
```
bibliotechy: I would love for the two places where this is being applied to be replaced with a method that describes what it is doing.

PR #182, opened 2018-01-04
Build QA site for PR

dkinzer:
REF BL-173

Description
===

Create auto QA builds from PR branches.

QA Steps
===
- [x] Site gets generated on PR request
- [x] Site gets attached to PR
- [x] Site gets updated on commit pushes to PR branch
- [ ] Site gets destroyed on PR close.

### DIFF
--- a/.docker/app/Dockerfile
+++ b/.docker/app/Dockerfile
@@ -1,7 +1,10 @@
 FROM ruby:2.4.1
 RUN apt-get update -qq && apt-get install -y build-essential libpq-dev nodejs
-RUN mkdir /tul_cob
-WORKDIR /tul_cob
+RUN mkdir /app
+WORKDIR /app
 ADD Gemfile .
 ADD Gemfile.lock .
 RUN bundle install
+COPY . .
+
+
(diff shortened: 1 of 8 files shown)


dkinzer:
First thing to note about this PR is that merging will not cause any issues for users (unless they use docker, but I think I'm the only one that does and I plan to document the impact soon).  So if this gets merged it will not break the site.  But, it will enable new PRs with this changed to have automatic QA sites built with links to those QA sites.

I should probably add that technically the PRs can have automatic QA sites generated for them with the flip of a switch on the jenkins server, but those sites will be broken because the changes in this PR makes it so that they wont be.

That said, to test the PR the following will need to take place.

* First you need to reenable the jenkins jobs the triggers the build: https://lurch.library.temple.edu/jenkins/job/Trigger%20PR%20QA%20Build/
image


* Next you willl need to make a PR off of this branch.

* When the PR gets generated you will see something like the following get added to the PR checks tab:
image
The message will change depending on what state the build is at.

* Once the build is complete it will either pass and the link should take you to a version of the site, or it will fail, the the link will point to the build console on jenkins so you can see what happened.

* Assuming the PR passed you can either merge the change or close the PR.  The staged build should go away so following the old link should not work anymore.

PR #185, opened 2018-01-10
Initial commit of solr records harvest test utility for review.

relaxing:


dkinzer:
I think there is a lot going on in this code as compact as it is.  So even though it looks short I found myself having to read it very carefully in order to understand what is going on.

There is some complexity that is coming from simultaneously dealing with harvesting records and the creation of the xm file.  I think that separating out those two concerns will go far into helping this code be easier to understand.

As it stands the code does not easily explain itself, so in lieu of that I think some general documentation at the top of the script explaining what the code is doing will help orient/prime the reader before they delve into the code itself.

relaxing:
@skng5 Since I all I want is to take the XML text verbatim from Solr and store it in a file, it made sense to me to not go through an intermediate step of parsing it. Do you think would I gain anything from deserializing the text, then serializing it again to file? I'm not trying to access the data or manipulate it in any way.

@dkinzer Fair point, I was thinking it could use comments as well :)

PR #187, opened 2018-01-12
Refactor specs

sensei100:
Lots of changes in the spec files here.
- Add the rubocop gem and changed spec files to get rid of syntax offenses
- Uncomment methods in the spec_helper file
- Cleans up code by removing explicit statements of spec type,  be_methodname rather that expressing a boolean value, using capybara negative selectors 
-Add context to conditional specs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,5 @@ coverage
 config/alma.yml
 config/bento.yml
 config/deploy_to.yml
+
+spec/examples.txt
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -133,3 +133,6 @@ Lint/EndAlignment:
 # Use my_method(my_arg) not my_method( my_arg ) or my_method my_arg.
 Lint/RequireParentheses:
   Enabled: true
+
+Rails:
+  Enabled: true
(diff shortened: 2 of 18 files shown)


sensei100:
I'm not sure that I'm the one to make a decision about adding rubocop to travis.  @bibliotechy  What do you think?

bibliotechy:
👍

sensei100:
Okay, I haven't worked with Travis configurations, so I will need to look into how to do that.  Right now, when I run rubocop there are 75 offenses in our code.  Does that mean that all those need to be corrected before Travis will pass?

bibliotechy:
I'd say let's add the config to make this run on travis, merge this and then afterwards we can apply the changes I just made in  https://github.com/tulibraries/tul_cob/pull/190 which should take care of our now failing builds.

dkinzer:
@sensei100 here is a an example of of how to do this on travis.  It's a one liner under the script: tag https://github.com/tulibraries/roper/blob/master/.travis.yml#L8

sensei100:
Since there is another PR for adding Rubocop,  I think this one is ready to merge.

PR #189, opened 2018-01-15
Bl 244 update home page

sensei100:
BL-244 This updates the text that displays on the home page.  It also uses the blacklight has_search_paramters? method to limit the facets displayed on the home page.  
- The home page should show only the facets for availability, library, and resource type.
screen shot 2018-01-15 at 9 23 56 am
- When search terms are used, all facets should render.
screen shot 2018-01-15 at 9 24 55 am

### DIFF
--- a/app/helpers/blacklight_helper.rb
+++ b/app/helpers/blacklight_helper.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+module BlacklightHelper
+  include Blacklight::BlacklightHelperBehavior
+
+  def only_home_facets(solr_parameters, user_paramters)
+    solr_parameters["facet.field"], solr_parameters["facet.pivot"] = home_facets, [] unless has_search_parameters?
+  end
+end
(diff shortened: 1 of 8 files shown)


app/helpers/blacklight_helper.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module BlacklightHelper
+  include Blacklight::BlacklightHelperBehavior
+
+  def only_home_facets(solr_parameters, user_paramters)
```
bibliotechy: Where does this get used?

bibliotechy:
So, this PR does exactly what it is supposed to, but I wonder if we could actually do a little bit better. Right now, we are asking Solr to generate the all facets on the home page, but we are not actually using them. So Solr is performing lots of unnecessary processing every time someone visits the home page.  If we look at the [home page as json](http://localhost:3000/?format=json) it is returning 11 facets even though we are only displaying three. I wonder if we could tell solr to only process the facets we want?
Feels like that should be a separate ticket, performance improvement, that should not block this from being merged.

sensei100:
The method created in the blacklight helper didn't end up getting used, so I removed that from the PR.

bibliotechy:
👍

PR #198, opened 2018-01-18
Update trim punctuation

sensei100:
BL-230 Testing revealed that trailing periods preceded by a parentheses were not being removed.  
- overrides traject's trim_punctuation macro for creator_facets
- updates the creator_name_trim_punctuation method so that names on record pages link to the facet
- adds specs to test several different punctuation contexts
- in the spec/marc_fixture.xml file record 991012024509703811, you should expect the period to be visible in the middle initial of Whitney, Ellen M. and removed from 4 Learning(Firm)
screen shot 2018-01-18 at 12 15 47 pm

### DIFF
--- a/app/models/traject_indexer.rb
+++ b/app/models/traject_indexer.rb
@@ -16,6 +16,10 @@ solr_url = ERB.new(solr_config["url"]).result
 require "traject/macros/marc21_semantics"
 extend  Traject::Macros::Marc21Semantics
 
+# Overrides the trim_punctuation method to remove periods preceded by parentheses
+require "traject/macros/custom_marc21"
+
+
 # To have access to the traject marc format/carrier classifier
 require "traject/macros/marc_format_classifier"
 extend Traject::Macros::MarcFormats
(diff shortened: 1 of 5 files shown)


lib/traject/macros/custom_marc21.rb
```diff
@@ -0,0 +1,49 @@
...
+
+      # trailing period if it is preceded by at least three letters (possibly preceded and followed by whitespace)
+      str = str.sub(/( *[[:word:]]{3,})\. *\Z/, '\1')
+
+      # single square bracket characters if they are the start and/or end
+      #   chars and there are no internal square brackets.
+      str = str.sub(/\A\[?([^\[\]]+)\]?\Z/, '\1')
+
+      # removes period when preceded by a paentheses
+      str = str.sub(/(?<=\))\./ , "")
```
dkinzer: Cool, I just learned something new about regex.  I did not know you could describe the subject that way.

app/models/traject_indexer.rb
```diff
@@ -16,6 +16,10 @@
 require "traject/macros/marc21_semantics"
 extend  Traject::Macros::Marc21Semantics
 
+# Overrides the trim_punctuation method to remove periods preceded by parentheses
```
dkinzer: A comment like this is probably better situated over the method that is being overridden.  It's very likely be be completely missed here (it's not a big deal in this case because that method is meant to be overridden [imho]).

PR #199, opened 2018-01-18
Override blacklight citation modal tool in views/catalog.

cdoyle-temple:
Requested by Emily: Jira #BL-240

(also added missing Close button to modal)


sensei100:
Looks much better with the padding and the close button makes for a better user experience.

PR #201, opened 2018-01-19
Fix first boolean operator dropped.

dkinzer:
REF BL-253

The `parse_queries` method used by some of our search_builder processing
is erroneoulsy dropping boolean operators from the first query.

This change refactors that method so that no information is lost.

 ## QA Steps
 * In libqa or locally run an advanced search query where the first query is "contains" "NOT moon" and the second query is connected with (*) AND  "contains" full.  Note the results.
 * Pull in the code and run the same query
 - [ ] Verify that the new search is limited as expected.

### DIFF
--- a/app/models/search_builder.rb
+++ b/app/models/search_builder.rb
@@ -46,14 +46,18 @@ class SearchBuilder < Blacklight::SearchBuilder
         # We de-reference values in order to be able to quote them; otherwise,
         # solr throws a 500 error: https://stackoverflow.com/a/10183238/256854
         #
-        # First we generate something like:
-        # [[["{}", "foo", "AND"], "q_1"], [["{}", "bar", "OR"], "q_2"]]
-        # And then we map that to ["_query_:..", "_query_:..."]
+        # First we take the query string and generate something like:
+        # [[["AND _query_:", "foo", "bar"], "q_1"], [["OR _query_:", "biz", "buz"], "q_2"]]
+        # And then we transform and rejoin that into a dereferenced query:
+        #
+        # "AND _query_:\"{foo v=$q_1}\" OR _query_:\"{biz v=$q_2}\""
+        #
         # @see :parse_queries, and :param_dereference methods below for more
         # details.
         queries = parse_queries(solr_parameters["q"])
           .zip(fields.keys)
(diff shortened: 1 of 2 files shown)


bibliotechy:
I was not able to replicate the expected behavior. Using the test fixture data only, I performed a search for:
all fields | contains | NOT Whiteman
and
all fields | contains | Philadelphia

I expected " A Philadelphia perspective : the diary of Sidney George Fisher covering the years, 1834-1871 / edited by Nicholas B. Wainwright." to not be present, but it was present.

However, if I perform

all fields | contains | Philadelphia
not
all fields | contains | Whiteman

I get the expected result of " A Philadelphia perspective : the diary of Sidney George Fisher covering the years, 1834-1871 / edited by Nicholas B. Wainwright." not being present.

dkinzer:
ok, i'll take a look at it once i'm done with BL-247

dkinzer:
> I was not able to replicate the expected behavior. Using the test fixture data only, I performed a search for:
all fields | contains | NOT Whiteman

@bibliotechy I think that because you are searching "all fields" it's returning records when any field satisfies the "NOT Whiteman" criteria.  Thus, this might explain your observation.

bibliotechy:
OK, limited to a single field and same result:

title | contains| Handbook
and
title | contains | NOT Genetic

But "Handbook of genetic skin disorders"  is listed.

However, with:

title | contains | Handbook
NOT
title | contains | Genetic

"Handbook of genetic skin disorders"  does not show up in the results list.

dkinzer:
> title | contains| Handbook
> and
> title | contains | NOT Genetic

Can you please test with 
title | contains| NOT Genetic
and
title | contains |  handbook

The issue is with how search is handled by the first item not the rest.  I don't know if advanced query updates the other queries the way it does the first one.

bibliotechy:
Same result

dkinzer:
ok, thanks.  I think I see the issue.  I't looking for "NOT,  'NOT Handbook'" . :) I'll need to also update the passed in value.

dkinzer:
ok, I just pushed a commit that removes the prefix boolean from the search term.

PR #203, opened 2018-01-19
List in descending chronological order

skng5:
Will handle overlapping and duplicate date ranges in separate PRs.


skng5:
Also needs to handle overlapping and duplicate ranges

lib/traject/macros/custom.rb
```diff
@@ -123,7 +123,7 @@ def extract_electronic_resource
           end
           # Sort on availability
```
dkinzer: Would it be possible to move this section to it's own function? The reason being that the current function is about extraction and this block is about sorting.  The function will get unnecessarily complicated because it's handling two different concerns.

The traject indexing DSL already allows for passing multiple processing functions into the `to_field` method. 

May I suggest we update the traject setting to be
```
to_field "electronic_resource_display", extract_electronic_resource, sort_electronic_resorce
```
And then we can define `sort_electronic_resource` and test it separately from the extraction functionality.

dkinzer:
I tried rerunning the travis test and it failed again.  I can't figure out why it's failing.  The obvious is not that helpful.

dkinzer:
oh, I just didn't wait long enough for the log to catch up.  It's failing on a test:

```
 1) Traject::Macros::Custom#extract_electronic_resource Only PRT fields are present multiple PRT fields present maps a multiple PRT fields to electronic_resource_display
     Failure/Error:
       expect(subject.map_record(records[2])).to eq(
         "electronic_resource_display" => ["foo", "bar"]
       )
     
       expected: {"electronic_resource_display"=>["foo", "bar"]}
            got: {"electronic_resource_display"=>["bar", "foo"]}
     
       (compared using ==)
     
       Diff:
       @@ -1,2 +1,2 @@
       -"electronic_resource_display" => ["foo", "bar"],
       +"electronic_resource_display" => ["bar", "foo"],
```

dkinzer:
Ah, this is the issue I was referring to in my comment (though I didn't know it would cause an actual test failure).  I already created tests for this function but now they are failing because the function is also performing sorts.  If we separate out the sorting into it's own function then not only will the test pass again, but both extraction and sorting can be kept easier to understand and maintain.

skng5:
@dkinzer As far as the test failing, that's what I was addressing in my original comment.

    Will handle overlapping and duplicate date ranges in separate PRs.

Since this is failing travis and the PR is still open, I'll address this and the separation of functionality in this branch.

skng5:
@dkinzer @bibliotechy Ready for review. Still needs tests. I'll file an issue for that.

dkinzer:
@skng5 do you mind removing the new test fixtures until you actually add new tests, thanks.

skng5:
They were not easy to collect. Do we have a durable place to keep them in the mean time besides my local computers?

skng5:
@dkinzer I'll remove it from the PR and will create new fixture when tests are ready.

spec/fixtures/marc_files/url_field_examples.xml
```diff
@@ -94,4 +94,55 @@
       <subfield code="u">http://foobar.com</subfield>
     </datafield>
   </record>
+  <record>
+    <!-- 9. Multiple PRT field with single non-overlapping coverage dates  -->
```
dkinzer: This extra fixtures aren't being used they should be removed until they are needed

skng5:
Known bug: If primary and/or secondary sort fields match, tertiary sort field renders in reverse order, which is not what we want. This will be addressed in a separate issue.

lib/traject/macros/custom.rb
```diff
@@ -145,6 +137,28 @@ def extract_electronic_resource
...
+            available = []
+          end
+          [available[1] || "9999", available[2] || "9999", title, subtitle || ""]
+        }.reverse!
+      end
+
+      def sort_electronic_resource
+        lambda do |rec, acc, context|
+          # Sort on availability
+          unless acc.empty?
```
dkinzer: `unless ac.empty?` here causes this function to return nil in the case that accumulator is empty.

lib/traject/macros/custom.rb
```diff
@@ -145,6 +137,28 @@ def extract_electronic_resource
         end
       end
 
+      def sort_by_coverage(acc)
+        acc.sort_by! { |r|
```
dkinzer: This function works by mutating the array in place which is cool, but the method name does not reflect that we can add a ! to the method to reflect this.

dkinzer:
This PR against this branch addresses my concerns, ... feel free to merge it to this one: https://github.com/tulibraries/tul_cob/pull/221

lib/traject/macros/custom.rb
```diff
@@ -145,6 +137,26 @@ def extract_electronic_resource
...
+            acc.sort_by! { |r|
+              subfields = r.split("|")
+              available = /Available from (\d+).* until (\d+)/.match(subfields.last)
+              title = subfields[1]
+              subtitle = subfields[2]
+              unless available
+                available = []
+              end
+              [available[1] || "9999", available[2] || "9999", title, subtitle || ""]
+            }.reverse!
```
dkinzer, line 152: This reverse! is causing BL-300. Not sure why the reverse! was added here when Orig ticket never mentioned anything about reversion the order.

PR #204, opened 2018-01-19
Fix search with colons fail.

dkinzer:
## Description

REF BL-252

Searches with : in them fail because : is a special field operator character in solr that modifies how the query operates.  Since our users do not know the names of any of our fields, it does not make sense to have this feature enabled.  Otherwise, it is just an unnecessary hurdle for our users.

In order to disable the feature, we need to substitute the colons with spaces. Escaping the character
doesn't work. 

## Dependencies
* Depends on #201, review and merge that first.

## QA Steps
* Before pulling down the code try searching for "Religious liberty in : America  the First Amendment in historical and contemporary perspective / Bruce T. Murray." locally.  You should get nothing.
- [ ] After pulling down the code verify that the same search pulls one record.

### DIFF
--- a/app/models/search_builder.rb
+++ b/app/models/search_builder.rb
@@ -34,17 +34,15 @@ class SearchBuilder < Blacklight::SearchBuilder
   end
 
   def escape_colons(solr_parameters)
-    queries = parse_queries(solr_parameters["q"])
+    query = solr_parameters["q"] || ""
 
-    # [ c, l, q ] = [ connector, local_param, query ]
-    solr_parameters["q"] = queries
-      .map { |q| c, l, q = q; [c, l, q.gsub(":", "\\:")] }
-      .map(&:join)
-      .join(" ")
+    return unless !query.empty?
 
-    # In the advanced search context the dereferenced values must be escaped.
+    # In the advanced search context dereferenced values must be escaped.
     if blacklight_params["search_field"] == "advanced"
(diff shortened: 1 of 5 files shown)


dkinzer:
:( It's going to be merge conflicts all the way up because merge squashes lose merging info.

bibliotechy:
Yeah, can happen when you squash but only if you build new feature branches off existing PRs instead of off master.

dkinzer:
@bibliotechy true that, but I built off this PR intentionally because each change was on the same area of code.

bibliotechy:
This fixes the problem in regular search, and in Advanced search with All Fields. 

When limited to title in advanced, it is still a problem, but I think that is out of scope for this issue.

I'm going to add some feature tests just in case the issue is affected by any of our future changes to our indexing. Just in case.

dkinzer:
I'm surprised (but believe you) that the advanced scenario isn't working since I substitute the : even in the advanced case.

In the follow up ticket I really want to take a close look at exactly how the advanced_search plugin works.  I think it's doing some changes that I'm not expecting.

dkinzer:
One assumption I've been making that may prove false is that the advanced_search plugin does not employ the same dereferencing trick that I use to do the begins_with and contains search.  If it does under certain conditions, I could be double referencing values.

bibliotechy:
@dkinzer I wouldn't put too much more time into this. I think colons in title field in advanced search is an edge case we can afford to let slide for a bit. It works from main search, and I think that is the most crucial thing.

PR #206, opened 2018-01-22
Normalize &/and queries

dkinzer:
REF BL-267
 
 ## Description
 Normalizes &/and searches by substituting " & " occurances in queries with
 "and".

 ## Dependicies
 * Review and merge #204 before merging this PR.

 ## QA Steps
 * Build locally.
 * Run a search that has & in it.
 - [ ] Verify that the same search with "and" instead of & has the same results.


dkinzer:
closing in favor of some version of #216

PR #207, opened 2018-01-22
Bl 278 book covers

sensei100:
Add book cover images to record show pages
screen shot 2018-01-22 at 3 40 37 pm


spec/views/catalog/_show_default.html.erb._spec.rb
```diff
@@ -12,7 +12,9 @@
     @config = Blacklight::Configuration.new {}
     @context = Blacklight::Configuration::Context.new(controller)
     @document = SolrDocument.new(id: 1)
+    @document[:format] = []
     without_partial_double_verification do
+      allow(view).to receive(:blacklight_config).and_return(@config)
```
dkinzer: Looks like this line is a dupe of the one that follows it.

sensei100: Thanks for catching that, I took it out and pushed another commit.

PR #210, opened 2018-01-23
Limit facet query when not required.

dkinzer:
REF BL-247

 ## Description
 Not all queries require solr to return results for all the facets.
This change limits the facet fields queried under specific conditions.

 ## Dependency
 * Review and merge #206 first 

 ## QA Steps
 - [ ]  Verify that tests limit facets as describe
 - [ ] Do some searches and verify that facets continue to work as expected.


app/models/search_builder.rb
```diff
@@ -46,6 +46,17 @@ def substitute_colons(solr_parameters)
     end
   end
 
+  def limit_facets(solr_parameters)
+    path = "#{blacklight_params["controller"]}/#{blacklight_params["action"]}"
+    count = blacklight_params.count
+
+    if path == "catalog/index" && count == 2
```
bibliotechy: `count == 2` feels brittle. I added `?format=json` to look at the JSON being returned and I thought it wasn't working because it was returning all of the facets. Not sure there is another way to do this though. Hmmm.

dkinzer: I don't know if there is another way to know where we are from this model either.  The Rails `request`  method is not available (maybe because that's just a controller thing).

And, I couldn't see a distinction between queries from catalog/index prior to doing a search and catalog/index when doing a search that was different other than that prior to doing the search the blacklight_params is a minimal set of just the controller and action i.e. the count is == 2.

Queries directly to catalog/*.json for some reason bipass this hook altogether.

bibliotechy: Yeah, ok. Maybe just add a comment about why we are checking for 2 blacklight params to help us remember in the future?

dkinzer: got it, and done :)

PR #211, opened 2018-01-23
This should be everything needed to perform the known item query tests.

relaxing:
It also addresses @dkinzer 's comment about lack of comments in solr_harvest_json_records2xml :)


bin/solr_harvest_json_records2xml
```diff
@@ -3,40 +3,65 @@ require 'net/http'
...
+    http = Net::HTTP.new(uri.host, uri.port)
+    http.use_ssl = true
+    http.verify_mode = OpenSSL::SSL::VERIFY_NONE
+    request = Net::HTTP::Get.new(uri.request_uri)
+    response = http.request(request)
+    page = JSON.parse(response.body)["response"]
 
+    # for every item on the page of results...
     page["docs"].each do |doc|
-      recordid = doc["id"]
```
bibliotechy: this line was deleted, but `recordid` gets used on line #29

lib/tasks/tulsearch.rake
```diff
@@ -7,7 +7,11 @@ namespace :fortytu do
 
     desc "Posts fixtures to Solr"
     task :load_fixtures do
-      Rake::Task["ingest"].invoke("spec/fixtures/marc_fixture.xml")
+      Dir.glob("spec/fixtures/*.xml").each do |file|
```
dkinzer: `Dir.glob` result order is not guaranteed to be the same in all systems possibly causing issues for processes that are not commutative or idempotent.

spec/features/searches_spec.rb
```diff
@@ -14,7 +14,7 @@
       visit "/"
       fill_in "q", with: item["title"]
       click_button "Search"
-      within(".document-position-0 h3") do
+      within first(".document-position-0 h3", text: item["title"]) do
         expect(page).to have_text item["title"]
```
dkinzer: I could not find any examples of `expect` inside of a `within` construct in the Capybara docs, possibly because `within` messes with expects natural waiting magic.  I believe `expect` should last.  Also `expect` can take arguments like `find` or `first` similar to `within`.

spec/fixtures/marc_fixture.xml
```diff
@@ -25286,13 +25286,13 @@
       <subfield code="b">.A66</subfield>
     </datafield>
     <datafield tag="210" ind1="0" ind2=" ">
-      <subfield code="a">Annu. rev. public health</subfield>
+      <subfield code="a">Annu. rev. SpaceDoDoPistols</subfield>
```
dkinzer: Some of the records in these examples are getting clobbered on travis by records that share the same id as records in `spec/fixtures/test_search_marc_records.xml`

dkinzer:
PR #225 against this branch applies the changes I am looking for.

lib/tasks/tulsearch.rake
relaxing: Dang. Great catch.

spec/features/searches_spec.rb
relaxing: I'm not sure what the issue is here? I don't care either way, but I don't believe this code was at fault. Can you clarify?

dkinzer: It's just a guess.  When I read the docs on capybara's automatic reload behavior in `expect` I decided that there is a potential for asynchronous issues to pop up if it's not used correctly. 

I noticed that there are no examples on the github page for capybara with `expect` within a `within` construct.  I took this to be an indicator that this was probably not expected.

Just now I googled to see if anyone has ever noticed an issue with `expect` inside of `within` and I only found this post from a while back http://atodorov.org/blog/2016/04/24/capybaras-within-altering-expectpage-scope/ , the developer here gets around his error by creating a new session.  Which I guess might be another way to deal with the issue I think this will have, but I don't know.

PR #216, opened 2018-01-24
Map & to and via synonym.

dkinzer:
REF BL-267

Adds a synonym map for & to and.

## Depends on
* #211


dkinzer:
Note: this is incomplete until we either update fieldtype "text" to use synonyms or we update our fields of type "text" to be "text_en" which already has synonyms configured at query time.

We may still reconsider this approach if we cannot apply it generally to all our fields.

PR #222, opened 2018-01-26
Search results

cdoyle-temple:
Implements styles from Rachel's mockups for search results page.


sensei100:
The only issue I see is the facets_helper override.  We have a file in the helpers directory  called facets_helper.rb that includes the Blacklight::FacetsHelperBehavior.  I think you should move the methods that you changed into this file rather than overriding the whole Blacklight facet helper. It looks like there are also some merge conflicts that need to be fixed now.

PR #223, opened 2018-01-26
Fix dperecation warning on count.

dkinzer:
Fixes a bunch of deprecation warnings on this line when testing.

### DIFF
--- a/spec/features/home_facets_spec.rb
+++ b/spec/features/home_facets_spec.rb
@@ -6,10 +6,11 @@ RSpec.feature "Advanced Search" do
 
   describe "Facets rendered" do
     it "Only a subset of all the facets render on the homepage" do
+      visit "/catalog?search_field=all_fields&q_1=x"
+      search_facets = page.all(".facet_limit").length
+
       visit "/catalog"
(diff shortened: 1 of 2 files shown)


spec/features/home_facets_spec.rb
```diff
@@ -6,10 +6,11 @@
 
   describe "Facets rendered" do
     it "Only a subset of all the facets render on the homepage" do
+      visit "/catalog?search_field=all_fields&q_1=x"
```
bibliotechy: What does this have to do with the deprecation?

dkinzer: Nothing, but I just started to randomly get an error on this test.  The search was pulling in 3 facets even though it should be 10.  I think because it was not done with the search (it's before I did all my reading on capybara docs).

Anyway, currently these tests seem to be brittle to timing issues.

dkinzer: btw, I was planning to do a PR to make this less brittle to timing problems after this gets merged, but I gess I can do that now and just reset this change... give me a sec.

PR #228, opened 2018-01-29
Make spec less brittle to timing issues.

dkinzer:
By adding a `have_text` expectation we force capybara to retry
for up to 2 seconds (by default).  This makes sure that we get the
number of facets on the rendered search page instead of potentially on
the front page where it will be 3 thus creating a false negative when
3 is not < 3.

### DIFF
--- a/spec/features/home_facets_spec.rb
+++ b/spec/features/home_facets_spec.rb
@@ -11,7 +11,6 @@ RSpec.feature "Advanced Search" do
 
       visit "/catalog?search_field=all_fields&q=test"
       expect(page).to have_current_path("/catalog?search_field=all_fields&q=test")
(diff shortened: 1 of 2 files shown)


dkinzer:
This is what I was talking about regarding timing issues... you can see that the page has not actually changed over so that's why it's failing.  I'm going to update the default wait time to 5 seconds vs 10.  But I would still like to keep this change in place because the failure at this point shows that it's because we have not switched to the new page.

dkinzer:
Weird, looks like it keeps getting stuck on the front page or the wait feature is not working as advertised.  I'm going to just let this be for now until I get some time to read more about how to properly handle timing issues in capybara specs.

PR #230, opened 2018-01-29
Bl 226 show unable to renew indicator

skng5:
Issue #BL-226


skng5:
None of the actual code changed, need to examine the data we receive from Alma.

PR #239, opened 2018-01-31
Bl 226 show unable to renew indicator

skng5:
The data viewed had no items that were not renewable. Run test against user with uid: "20180130" to see one item that is not renewable.

Spec for view partial added to verify logic non-renewable indicator logic.


bibliotechy:
LOve it. Just add tests.

PR #244, opened 2018-01-31
Add arrow functionality to not available buttons

sensei100:
Adds dropdown arrow to items that are "not available"


bibliotechy:
When I tested this locally (on Firefox) and the buttons look like
buttons-avail

However in Chrome it looks like:
chrome

sensei100:
I adjusted the css white-space property and it looks right in chrome, safari, and Firefox.

PR #249, opened 2018-02-01
Update search results theme/layout based on new mockup

cdoyle-temple:
REF BL-276


app/views/catalog/_previous_next_doc.html.erb
```diff
@@ -0,0 +1,25 @@
+<% #Using the Bootstrap Pagination class  -%>
+<% #DEPRECATED - using id="previousNextDocument" as a selector is deprecated and will be removed in Blacklight 6.0 %>
```
dkinzer: This deprecation warning is a little scary, wondering if we should be doing something else here.

PR #253, opened 2018-02-01
- Updates to query results test and test data

relaxing:
- Add'l documentation to test code
- Fixes for harvester not handling MMSID search
- Woah-oh, amber is the color of my energy

### DIFF
--- a/spec/controllers/query_results_spec.rb
+++ b/spec/controllers/query_results_spec.rb
@@ -11,36 +11,47 @@ RSpec.describe CatalogController, type: :controller do
       fixtures = YAML.load_file("#{fixture_path}/search_features.yml")
       test_queries = fixtures.fetch("results_queries")
       docs = []
+      # Take the test queries from the YAML file one at a time...
       test_queries.each do |test_item|
         search_string = ""
-
+        # Concatenate the specified search terms into one search string
         test_item["query_type"].each do |query_field|
           search_term = test_item[query_field]
           if search_term.kind_of?(Array) == true
-            search_term = search_term.to_s
+            search_term = search_term[0].to_s
           end
           search_string += search_term + " "
         end
 
-        it "for search #{search_string}" do
+        it "for search #{search_string} contains the correct items in the correct order" do
+          # Perform the search and concatenate the first two pages of results (we aren't interested in results past 2 pages)
           for page in [1..2] do
             get(:index, params: { q: search_string, page: page.to_s }, format: "json")
             docs += JSON.parse(response.body)["response"]["docs"]
           end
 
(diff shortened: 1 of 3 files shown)


spec/controllers/query_results_spec.rb
```diff
@@ -11,36 +11,47 @@
...
+              # ...And if the item is found in our list...
               if test_item["doc_id"].include?(search_result["id"])
+                # ...Remove it from the list of expected items
+                # The goal being to empty the list of expected items.
+                # Once it's empty, we know all items have been found and the test is fulfilled.
                 test_item["doc_id"].delete(search_result["id"])
               end
-            elsif test_item.include?("secondary_id") && !(test_item["secondary_id"].nil?) && test_item["secondary_id"].length > 0
+            # Once we've emptied the list of primary results, now we can look for secondary results in the same manner
+            elsif test_item["doc_id"].size == 0 && test_item.include?("secondary_id") && !(test_item["secondary_id"].nil?) && test_item["secondary_id"].length > 0
```
dkinzer: The comments are great but I stilI find the use of `size == 0` in these tests to be confusing.  RSpec has an `include` matcher that is more descriptive and does not require you do to all this deleting preparation that adds to the confusion (imho).

Ultimately you want to verify that a subset of results is included in a super set.  See the following example of how that can be accomplished using the `include` matcher.

```
RSpec.describe "How to test inclusivity" do
  context "testing a subset of a set" do
    it "should work to test include with spread operator" do
      a = [1, 3, 2, 7, 5]
      b = [ 3, 7]

      expect(a).to include(*b)
    end
  end
end
```

relaxing: @dkinzer Testing include() is not sufficient, because we are also concerned with order, not just existence.

dkinzer: @relaxing 😎 that's important, but it's not obvious to me how you are accounting for order.  Is it possible to add something to make that a bit more obvious since it will probably come up again in the future.  Reading the code I see `test_item["doc_id"].delete(search_result["id"])`, but as far as I know `.delete` does not care about order.

Also if order is an issue why aren't we doing a straight up comparison between two arrays of ids? `[1, 2, 3] == [3, 1, 2]` vs `[1, 2, 3] == [1, 2, 3]` (first is false second is true).

relaxing: @dkinzer But we're not concerned with exact order, just that all the members of one set occur before any members of a second set. So that doesn't work either.

dkinzer: OK, I get it now.  Thanks!

PR #256, opened 2018-02-02
Update label for availability status

sensei100:
Updates status for checked out materials to include the library name.  
- A search for 991021718189703811 should display like this:
screen shot 2018-02-02 at 2 59 40 pm


dkinzer:
Loving the use of the higher order function here.

PR #257, opened 2018-02-02
Add image_tags to record pages

sensei100:
Add image tags for covers and default icons to the full record page.


dkinzer:
picture is not showing up... only a blank box: https://book_cover_record_page.devkinzer.net/catalog/991030023269703811

dkinzer:
looks great on my local, must have been an issue with the instance

PR #268, opened 2018-02-05
Accept alpha/num/space/parentheses

skng5:
Issue #310


skng5:
What other characters will we need to accept?

skng5:
Found cases of trailing spaces it's not catching.

dkinzer:
@skng5 I'm not a regex guru, can you please give me some examples of what the expected behavioris you are looking for:

For example  from IRB expected:
```
"????!!!David &&&!!! World".gsub(/[^[:alnum:])]*$/, "")
```

dkinzer:
Above results in the following output:

```
"????!!!David &&&!!! World"

```

dkinzer:
OK, I get it.  Anything that is not either an alpha numeral or ), and is at the end of a string gets deleted.  In that case, looks good.

PR #272, opened 2018-02-06
Fix shib logout not redirecting properly.

dkinzer:
REF BL-224

Redirects to shib url on logout.

QA Steps
===
- Log in as a user (you'll need to temporarily enable forms.)
- Click Log Out button
- [ ] Verify that you get a redirect to a shib URL.  Note you will get error on local because your local does not support shib.


bibliotechy:
I made a tiny change to this, defaulting to the production IDP for logout instead of the non production one. Once that passes, I'll merge. Looks good.

PR #275, opened 2018-02-07
BL-322 Update text for banner and landing page

sensei100:
Banner and home page text updated.  Link is not published yet, so you will see access denied when you click on it.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -9,16 +9,6 @@ module ApplicationHelper
     render_location(value[:value].first)
   end
 
-  def string_with_link(str, link_url, link_options = {})
-    match = str.match(/__([^_]{2,30})__/)
-    if !match.blank?
-      raw($` + link_to($1, link_url, link_options) + $')
-    else
-      raise "string_with_link: No place for __link__ given in #{str}" if Rails.env.test?
-      nil
-    end
-  end
-
   def get_search_params(field, query)
     case field
     when "subject_display"
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -9,6 +9,16 @@ module ApplicationHelper
     render_location(value[:value].first)
   end
 
+  def string_with_link(str, link_url, link_options = {})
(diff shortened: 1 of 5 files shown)


app/helpers/application_helper.rb
```diff
@@ -9,6 +9,16 @@ def render_location_show(value)  # why do we get the entire solr document in sho
     render_location(value[:value].first)
   end
 
+  def string_with_link(str, link_url, link_options = {})
```
dkinzer: Ruby has an idiomatic way to handle formatting strings using the `%` operator.

`puts "Hello you %s, %s world" % ["beautiful", "funny"]`

config/locales/blacklight.en.yml
```diff
@@ -1,8 +1,10 @@
 en:
   blacklight:
     application_name: "Library Search Beta"
-    greeting: "Try our new beta Library Search"
-    introduction: "Based on user feedback, Temple University Libraries are in the process of building an enhanced version of the Library Search. This is still under development, as we continue to add new features and functionality through spring 2018. Try out the new beta and let us know what you think at asktulibrary@temple.edu!"
+    greeting: "Try out the Library Search beta"
+    introduction: "An enhanced version of the Library Search is currently under development at Temple University Libraries. This new search interface is currently in beta and contains only records from our local collections (such as books, journals, and media). As work continues through the Spring 2018 semester, the Libraries will incrementally be adding new features and incorporating more records, including those for online articles."
+    introduction_html: "Please refer to our __Library Search Road Map__ for more information about work in progress and providing feedback."
```
dkinzer: This string should be: `"Please refer to our %s Search Road %s for more information about work in progress and providing feedback."`, maybe with a comment above for clarity on what each %s is for (there might be a way to label the %s bit too (that's possible in PHP not sure if it's possible in Ruby).

sensei100:
@dkinzer  I made the change you requested.  My thought with the helper method was that it would be reusable for other links we eventually put there, rather than having to assign the value each time.

sensei100:
Fixed the typo and added the section_heading class back to the p tag.

PR #283, opened 2018-02-24
Allow for reuse of downloaded files.

dkinzer:
REF BL-76

### DIFF
--- a/lib/tasks/oaiharvest.rake
+++ b/lib/tasks/oaiharvest.rake
@@ -45,6 +45,7 @@ namespace :fortytu do
       end
 
       Rake::Task["fortytu:oai:ingest_all"].invoke()
+      Rake::Task["fortytu:oai:purge"].invoke()
 
       # Check the build for errors.
       if File.file? "log/fortytu.log.error"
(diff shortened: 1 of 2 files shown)


lib/tasks/oaiharvest.rake
```diff
@@ -19,30 +19,35 @@ namespace :fortytu do
...
 
-      # Delete the previous build's marc_xml_files.
-      Dir.glob("tmp/alma/**/*.xml").each { |file| File.delete file }
-
       # Run the harvester.
-      Rake::Task["fortytu:oai:harvest"].invoke(from, to)
-      Rake::Task["fortytu:oai:conform_all"].invoke()
+      if !args[:use_cache]
+        # Delete the previous build's marc_xml_files.
+        Dir.glob("tmp/alma/**/*.xml").each { |file| File.delete file }
```
bibliotechy: There is blocking logic at a higher level that prevents two jobs from running concurrently? Otherwise, this could run into a problem where one process deletes another processes files.

dkinzer: @bibliotechy correct.  Jenkins jobs do not run concurrently by default: 
image

PR #286, opened 2018-03-01
Use default values in form

dkinzer:
REF BL-291
REF BL-334

The advanced search form is not currently using the default values for
the operators which are available in the blacklight_params.

This PR also addresses an issue where the op_row field on multiple searches without hitting the "restart" button in between the searches, was appending new values to the op_row array in the params object and then we were not using the newer values.


bibliotechy:
Glad to see that stacking values thing fixed!

PR #301, opened 2018-03-29
Use feature flags to turn off not quite ready features in production 

bibliotechy:


app/controllers/catalog_controller.rb
```diff
@@ -472,8 +472,8 @@ class CatalogController < ApplicationController
 
     # Do not show endnotes, SMS or email  for beta release
     config.show.document_actions.delete(:endnote)
-    config.show.document_actions.delete(:sms)
-    config.show.document_actions.delete(:email)
+    config.show.document_actions.delete(:sms) if Rails.configuration.features[:sms_document_action_disabled]
```
dkinzer: I think this will be easier to maintain if configuration  is two dimensional:

```
Rails.configuration.features.fetch(:document_action_disabled, []).each do |action|
  config.show.document_actions.delete(:sms)
end

# Then we only need to maintain the list of actions to disable in configuration vs remembering to update two places.
Rails.configuration.features[:document_action_disabled]= [ :sms, :email ]
```

config/initializers/features.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
```
dkinzer: Same comment as above, this is just a suggestion:
```
Rails.application.config.features[:disabled] = []
```
Thus, anything we add to this list would be true and we can iterate over the list to act on the contents.

app/controllers/catalog_controller.rb
bibliotechy: So, ubless I'm misunderstanding, that list would have to exist as ruby code, so it would mean different code in different environments, which is what we are trying to avoid. 

Toggling on ENV variables means it is very easy to add or remove the feature with no code changes.

dkinzer: Oh I thought `Rails.configuration` could fetch on specific environments so prod: vs dev: etc would be different... but yes, still in code.

```
prod:
  disabled:
     - :x  
     - :y 
     - :z
```

My point was more about being able to iterate over a config

bibliotechy: Well, our problem is that libqa and librarybeta both run in production environment. Development is more like local developers environment rather than what might be called staging

PR #302, opened 2018-03-29
Constraints

sensei100:
BL-319 Restores the field labels to constraints for advanced search
- The results for an advanced search should display the field name and search query term
- If and, or was used then there should be a constraint shown for each query term
- When not was selected, the "NOT" keyword should be displayed in the constraint as well
screen shot 2018-03-29 at 4 42 26 pm

### DIFF
--- a/app/helpers/advanced_helper.rb
+++ b/app/helpers/advanced_helper.rb
@@ -3,14 +3,6 @@
 module AdvancedHelper
   include BlacklightAdvancedSearch::AdvancedHelperBehavior
 
-  def label_tag_default_for(key)
-    if params[key].present?
-      params[key]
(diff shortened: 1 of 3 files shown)


dkinzer:
@sensei100 this seems to be reverting the code back to it's orig state, but that is removing some work I did to allow for extra flexibility in the advanced form.  From what we talked over on Friday it looks to me like maybe the issue is with the `search_field_def_for_key` method and if we fix that we should need revert it completely to it's original state.  Is that correct?

sensei100:
@dkinzer What is the extra flexibility that is missing?

dkinzer:
@sensei100 ...being able to add more more search constraints.. (i.e. it's not limited to three sets.)

PR #305, opened 2018-04-05
Bl 354 map locations

sensei100:
BL-354 Built a rake task to get location information from the retrieve locations api and use that to populate a yam file translation map for locations.  
- The location information under the library name should display as the value of the locations.yml file, rather than the key(for example: Stacks, not stacks)
screen shot 2018-04-05 at 9 05 09 am


ebtoner:
@sensei100 You can remove the location codes that include "delete," I believe. I think that this is left over from the Alma migration and shouldn't actually be attached to any item records that are unsuppressed.

PR #311, opened 2018-04-06
[WIP] Add range-limit query.

dkinzer:
REF BL-350

Depends on https://github.com/tulibraries/primo/pull/13

Adds range limit query date facet.


dkinzer:
Unfortunately the Primo/PNXS API does not currently handle range queries in an expected/documented fashion.  At this point it does not make sense to pursue this further.

PR #317, opened 2018-04-10
Add translations for primo facet codes.

dkinzer:
REF BL-360

* Add language translations: REF BL-363.
* Add availability translations: REF BL-364
* Add resource type translation REF BL-365


dkinzer:
@bibliotechy yes, primo_central_controller is deriving the lang facet from catalog_controller.  If I try to add this directly in primo_central_controller than blacklight throws an error saying that the lang facet has already been configured.

bibliotechy:
So...I've been thinking this over and I'm still uneasy about it. Let's talk through options tomorrow at the developers share, as a group.

dkinzer:
@bibliotechy it doesn't look like it's possible to share templates between two controllers without either explicitly using a shared template, or by deriving one controller over another.

Further, the explicit use of a shared template will fail if that template makes calls to non shared resources that aren't templated at the action level or have a default template in `views/default`.

I think our current approach of deriving the primo_central controller from the catalog_controller makes the most sense in terms of not duplicating templating efforts across multiple view templates.  However, it does come with the drawbacks you mention.

sul's approach is to duplicate its templates, as far as I can tell.  And that may be the right approach too.

A third option that I am looking into is to have both catalog and primo_central derive from another controller that is as yet configured (an abstract controller).  But I do not know how to make this happen as blacklight is already using the catalog  namespace...  But perhaps the is easily remedied by updating the controller set in config/routes.rb.

bibliotechy:
So when you call `config.add_show_field` in the `configure_blacklight` block, it is appending a config to an OrderedHash called `config.show_fields` containing hash items like
```ruby
"subject_display"=>
  #<Blacklight::Configuration::ShowField label="Subject", helper_method=:list_with_links, multi=true, key="subject_display", field="subject_display", if=true, unless=false>,
 "url_finding_aid_display"=>
  #<Blacklight::Configuration::ShowField label="Finding Aid", helper_method=:check_for_full_http_link, key="url_finding_aid_display", field="url_finding_aid_display", if=true, unless=false>,
...
```
So, maybe we just need a simple method that can override existing blacklight configs with the new params we want?

Something like
`config.override_show_field 'availability', helper_method: :my_new_helper_method`

PR #318, opened 2018-04-10
Bl 366 library display

sensei100:
screen shot 2018-04-10 at 3 35 57 pm
screen shot 2018-04-10 at 3 34 21 pm

BL-366 Library names should only display once, regardless of how many copies there are at that library
- Record 991025629919703811 should display "Paley Library" once
- Temporary libraries should display instead of library name if item is in a temporary library
- Record 991026548529703811 should display "Resource Sharing Library" NOT "Paley Library"

### DIFF
--- a/app/helpers/alma_data_helper.rb
+++ b/app/helpers/alma_data_helper.rb
@@ -43,8 +43,8 @@ module AlmaDataHelper
     end
   end
 
-  def library_name(item)
-    libraries = item.group_by do |lib|
+  def library_name(items)
+    items.group_by do |lib|
       (lib["holding_data"]["temp_library"]["value"] if lib["holding_data"]["in_temp_location"] == true) || (lib["item_data"]["library"]["value"])
     end
   end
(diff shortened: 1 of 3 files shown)


app/helpers/alma_data_helper.rb
```diff
@@ -43,14 +43,16 @@ def location_status(item)
     end
   end
 
-  def library_status(item)
-    if item["holding_data"]["in_temp_location"] == true
-      Rails.configuration.libraries[item["holding_data"]["temp_library"]["value"]]
-    else
-      Rails.configuration.libraries[item["item_data"]["library"]["value"]]
+  def library_name(item)
```
bibliotechy: I think the `item` parameter should be `items`. Future devs may be confused thinking only one item is being passed to this, but really it is one or more.

app/helpers/alma_data_helper.rb
```diff
@@ -43,14 +43,16 @@ def location_status(item)
...
+    libraries = item.group_by do |lib|
```
bibliotechy: We don't need the  `libraries =` here. Without it, the method will still return the value of `item.group_by do ... end `, and we don't have to do an assignment we never end up using.

sensei100:
@bibliotechy I made the changes you requested.

PR #325, opened 2018-04-13
Update alma data

sensei100:
Refines data being returned from the alma api
- BL-352 updates temporary locations logic to use value instead of description
- BL-372 redirects mapping to the traject .properties file and removes the extra .yml file


lib/tasks/locations.rake
```diff
@@ -20,7 +20,11 @@ namespace :locations do
       locations = location_response["locations"]["location"]
       location_list[code] =  {}
       locations.each do |l|
-        location_list[code][l["code"]] = l.fetch("external_name") if !l.fetch("external_name").nil?
+        if code == "RES_SHARE"
+          location_list[code][l["code"]] = l.fetch("name") if !l.fetch("name").nil?
```
dkinzer: l.fetch("name")` will only return nil if l has the key "name" and it is `nil` otherwise it will throw an error.

app/helpers/alma_data_helper.rb
```diff
@@ -38,9 +38,9 @@ def public_note(item)
   def location_status(item)
     if item["holding_data"]["in_temp_location"] == true
       if item["holding_data"]["temp_call_number"].empty?
-        "#{item["holding_data"]["temp_location"]["desc"]}"
+        "#{Rails.configuration.locations[item["holding_data"]["temp_library"]["value"]][item["holding_data"]["temp_location"]["value"]]}"
```
dkinzer: for the sake of readability, `temp_library` and `temp_holding` should be found separately and used as keys.

`Rails.configuration.locations[temp_library][temp_location]`

lib/tasks/locations.rake
dkinzer: Also it's more idiomatic to use `unless` in place of `if !`

sensei100:
@dkinzer I made the changes you requested.

PR #326, opened 2018-04-13
Clear article search constraints

skng5:


skng5:
It appears to behave correctly, but I'm unable to access the `Blacklight::Controller#search_action_path` from the tests. Suggestions?

PR #342, opened 2018-04-20
Suppress holdings

sensei100:
BL-373 Items with a process type of missing or lost should not display in the availability data. 
- Alma mms_id 991016705469703811 in the spec/fixtures/marc_fixture.xml file should display the copy that it available, but not the copy that is missing
screen shot 2018-04-20 at 2 46 32 pm

### DIFF
--- a/app/helpers/alma_data_helper.rb
+++ b/app/helpers/alma_data_helper.rb
@@ -52,10 +52,15 @@ module AlmaDataHelper
     end
   end
 
+  def missing_or_lost?(item)
+    item["item_data"]["process_type"].has_value?("MISSING") || item["item_data"]["process_type"].has_value?("LOST")
+  end
+
   def library_name(items)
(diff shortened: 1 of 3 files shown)


app/helpers/alma_data_helper.rb
```diff
@@ -53,7 +53,7 @@ def location_status(item)
   end
 
   def library_name(items)
-    items.group_by do |lib|
+    items.select { |item| item unless item["item_data"]["process_type"].has_value?("MISSING") || item["item_data"]["process_type"].has_value?("LOST") }.group_by do |lib|
```
bibliotechy: For readability, please break this up a bit. 

* A method helper called something like `missing_or_lost?(item)` that encapsulates the checks for Missing or Lost
* start the group by on the next line, indented.

It should look something like
```
 items.select { |item| item unless missing_or_lost?(item)}
  .group_by do |lib| 
...
```

sensei100:
@bibliotechy I made the changes you requested.

PR #344, opened 2018-04-21
Add "More" bento box.

dkinzer:
REF BL-378

Adds a bento box that grabs the format facet from the catalog
search and filters out "Book" and "Journal/Periodical" facets if
included:

image

Some other enhancements include.
* Does not search for facets in non facet bento searches
* Only searches for the facet.field = format for the more bento search.

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -10,7 +10,7 @@ class SearchController < CatalogController
     config.facet_fields = ActiveSupport::OrderedHash.new
     config.index_fields = ActiveSupport::OrderedHash.new
     config.sort_fields = ActiveSupport::OrderedHash.new
-    config.add_facet_field "format", label: "Resource Type", collapse: false, limit: false
+    config.add_facet_field "format", label: "Resource Type", url_method: :path_for_more_facet
   end
 
   @@per_page = 10
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -20,6 +20,7 @@ class SearchController < CatalogController
       searcher = BentoSearch::ConcurrentSearcher.new(*engines)
       searcher.search(params[:q], per_page: @@per_page, semantic_search_field: params[:field])
       @results = searcher.results
+      @response = @results["more"].first.custom_data
     end
   end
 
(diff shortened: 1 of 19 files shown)


app/controllers/search_controller.rb
```diff
@@ -1,13 +1,21 @@
 # frozen_string_literal: true
 
-class SearchController < ApplicationController
+class SearchController < CatalogController
```
bibliotechy: I'd love a little bit of clarity on why this inherits form Catalog Controller at this level? Just for the config?

app/models/search_builder.rb
```diff
@@ -4,6 +4,7 @@ class SearchBuilder < Blacklight::SearchBuilder
   include Blacklight::Solr::SearchBuilderBehavior
   include BlacklightAdvancedSearch::AdvancedSearchBuilder
   include BlacklightRangeLimit::RangeLimitBuilder
+  include BentoSearchBuilderBehavior
```
bibliotechy: All searcher need access to this, not just the Bento Searcher?

app/controllers/search_controller.rb
dkinzer: Now that we are asking this search_controller to render the format facet it's essentially acting like a variant of the search catalog.  Rendering that facet requires a lot of the functions that are already part of the catalog controller.  I began by first trying to include the specific modules that were needed just for the facet rending; however I kept needing to add more and more includes and then it must made more sense to inherit directly from CatalogController.

app/models/search_builder.rb
dkinzer: Since the bento searcher is now inheriting from catalog_controller I thought this was the easiest solution.  This behavior currently only adds some methods that gets used by the bento searcher.  However, the other option is to do what we did for primo_central and create a search builder behavior class specific to bento.... But this is heavy handed at this point in my opinion because unlike primo_central ... these search modifying methods are meant to update the solr query not another api specific query like in the case of primo_central.

This is the same strategy that blacklight_range_limit and blacklight_advanced_search gems use.

bibliotechy: :+1:

app/controllers/search_controller.rb
bibliotechy: Makes sense.

PR #345, opened 2018-04-23
Upgrade blacklight version to 7

dkinzer:
REF BL-399

Depends on:
* https://github.com/tulibraries/blacklight_alma/pull/11
* https://github.com/projectblacklight/blacklight_advanced_search/pull/83


dkinzer:
I'm not interested in going down this rabbit hole right now.  I did a little experimentation during the weekend that made me hopeful, but I think I just got lucky.  Now that I'm actually trying to implement it I'm running into one issue after another and it's just like little things to become unnecessary time sinks.

PR #348, opened 2018-04-24
Avoid deadlock in Bento seaches

bibliotechy:
Add eager loading and cache_classes changes to avoid concurrency deadlocks in Bento.

This required moving the traject config file out of app so it doesn't get eager loaded. All references to the file had to be updated.

Also had to remove an `included` statement from the primo `search_builder_behavior` and move it to the primo 'search_builder` directly.


dkinzer:
I keep not being able to set this up locally... image just dies:

```
xiting
/usr/local/bundle/gems/blacklight-6.14.1/lib/blacklight/configuration/fields.rb:133:in `add_blacklight_field': A facet_field with the key availability_facet already exists. (RuntimeError)
        from /usr/local/bundle/gems/blacklight-6.14.1/lib/blacklight/configuration/fields.rb:26:in `add_facet_field'
        from /tul_cob/app/controllers/catalog_proxy_controller.rb:215:in `block in <class:CatalogProxyController>'
        from /usr/local/bundle/gems/blacklight-6.14.1/lib/blacklight/configuration.rb:230:in `configure'
        from /usr/local/bundle/gems/blacklight-6.14.1/app/models/concerns/blacklight/configurable.rb:36:in `configure_blacklight
```

PR #350, opened 2018-04-25
BL-374 Add sort order for holdings

sensei100:
BL-374 Physical items should be sorted first by library name, then location, call number, and description

### DIFF
--- a/app/helpers/alma_data_helper.rb
+++ b/app/helpers/alma_data_helper.rb
@@ -56,7 +56,7 @@ module AlmaDataHelper
     item["item_data"]["process_type"].has_value?("MISSING") || item["item_data"]["process_type"].has_value?("LOST")
   end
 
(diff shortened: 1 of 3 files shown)


app/helpers/alma_data_helper.rb
```diff
@@ -57,19 +57,53 @@ def missing_or_lost?(item)
   end
 
   def library_name(items)
```
bibliotechy: @sensei100 One thought. I'm not sure `library_name` still describes what this method does. But it is a hard one to name since it does so many things. Maybe `group_and_order_items(item)`? What do you think?

sensei100:
@bibliotechy  I think it is a good idea.  I made the name change.

bibliotechy:
@sensei100 I think it needs to be changed in the view as well.

PR #352, opened 2018-04-26
Add direct link to availability if available.

dkinzer:
REF BL-395
REF BL-396

Refactors availability rendering for primo_central and allows for
a switch between either the iframe view or a direct link to the
resource.

### DIFF
--- a/app/models/blacklight/primo_central/document.rb
+++ b/app/models/blacklight/primo_central/document.rb
@@ -28,8 +28,7 @@ module Blacklight::PrimoCentral::Document
   end
 
   def has_direct_link?
-    availability = @_source.fetch("delivery", {})
-      .fetch("availability", [])
+    availability = @_source.dig("delivery", "availability") || []
     availability == ["fulltext_linktorsrc"]
   end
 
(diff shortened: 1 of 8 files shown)


app/models/blacklight/primo_central/document.rb
```diff
@@ -27,6 +27,12 @@ def initialize(doc, req = nil)
     super(doc, req)
   end
 
+  def has_direct_link?
+    availability = @_source.fetch("delivery", {})
```
bibliotechy: This seems lie a great opportunity to leverage dig, as noted [here](https://rollbar.com/blog/top-10-ruby-on-rails-errors/#2-nomethoderror-undefined-method--for-nilnilclass)

this could become
```ruby
availability = @_source.dig("delivery", "availability") || []
```
Whattya think?

bibliotechy:
Actually, one other thought...can a Primo record have more than on direct link?

dkinzer:
I have noticed that even when there are multiple links associated to a resource the GetIt link is always only one and it's either a direct link tot he resource an iframe with links.

app/models/blacklight/primo_central/document.rb
dkinzer: Yeah, that would work nicely.

PR #354, opened 2018-04-26
Fix issue with embedded content error for records with no policy data

sensei100:
There seems to be a situation where loan policy is not always provided.  
- Handles the case where policy has no data


bibliotechy:
Looks good!

PR #368, opened 2018-05-02
Make default format unknown.

dkinzer:
REF BL-411

Use the "unknown" format to avoid broken image issue.

### DIFF
--- a/app/helpers/catalog_helper.rb
+++ b/app/helpers/catalog_helper.rb
@@ -49,7 +49,12 @@ module CatalogHelper
       "text_resource" => "journal_periodical",
     }
 
-    format = default_image[format] || [ "unknown" ]
+    format = default_image[format]
+
+    if !File.exists?("app/assets/images/svg/#{format}.svg")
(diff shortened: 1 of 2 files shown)


app/helpers/catalog_helper.rb
```diff
@@ -45,10 +45,11 @@ def default_cover_image(document)
       "review" => "journal_periodical",
       "statistical_data_set" => "dataset",
       "technical_report" => "journal_periodical",
+      "book_chapter" => "book",
       "text_resource" => "journal_periodical",
     }
 
-    format = default_image[format] || format
+    format = default_image[format] || [ "unknown" ]
```
bibliotechy: This method needs to account for formats that have a standard svg in https://github.com/tulibraries/tul_cob/tree/master/app/assets/images/svg but are not included in the `default_image` hash. 

Calling `default_cover_image({format: "audio"})` needs to return `"svg/audio.svg"`, but this change would make it return `"svg/unknown.svg"`



```

PR #381, opened 2018-05-17
Replicate Bookmark feature for PCI data

dkinzer:
REF BL-414

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,5 +1,5 @@
 AllCops:
-  TargetRubyVersion: 2.2
+  TargetRubyVersion: 2.4
   # RuboCop has a bunch of cops enabled by default. This setting tells RuboCop
   # to ignore them, so only the ones explicitly set in this file are enabled.
   DisabledByDefault: true
--- a/Gemfile
+++ b/Gemfile
@@ -50,7 +50,7 @@ end
 # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
 gem "tzinfo-data", platforms: [:mingw, :mswin, :x64_mingw, :jruby]
 
-gem "blacklight"
+gem "blacklight", path: "blacklight"
 gem "blacklight_advanced_search"
 gem "blacklight-marc"
 gem "blacklight_range_limit"
@@ -91,7 +91,9 @@ gem "twilio-ruby"
 gem "skylight"
 gem "webpacker"
 gem "google-analytics-rails", "1.1.1"
-gem "primo", git: "https://github.com/tulibraries/primo"
+gem "primo",
+  path: "primo"
+  #git: "https://github.com/tulibraries/primo"
 gem "bootsnap"
 
 group :production do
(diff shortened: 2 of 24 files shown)


bibliotechy:
I'm getting an error when I try to view the bookmarks page.

Steps to reproduce
# Do a search for Solr results
# Mark two solr results as Bookmarks (Nav bar indicates 2 bookmarks)
# Do a search in Articles
# Select two articl results as bookmarks (Nav bar indicates 4 bookmarks)
# Click on Bookmarks link in navbar

```

ArgumentError in Bookmarks#index

Showing /home/chad/.rvm/gems/ruby-2.4.1@cob/gems/blacklight-6.15.0/app/views/catalog/_document_list.html.erb where line #3 raised:

'nil' is not an ActiveModel-compatible object. It must implement :to_partial_path.
----
Extracted source (around line #17): 

  <%= render partial: 'tools', locals: { document_list: @document_list } %>

  <%= render_document_index %>
  <%= render 'results_pagination' %>
<% end %>
</div>

----

Trace of template inclusion: app/views/bookmarks/index.html.erb
```

Not sure if I need to do something different to make this work as expected.

dkinzer:
@bibliotechy I can't reproduce the issue following the instructions. What's the specific search you are doing, maybe that's important.

Also can you please verify that you are on the latest code and that you have run `bundle install` as it needs to update the version of the primo gem we are using.

bibliotechy:
OK, now that I have it working, I have two noticed issues:

Alma availability doesn't seem to be working  - Maybe we just hide that for now?
screen shot 2018-05-24 at 2 59 27 pm

and a weird one - the per page is 2

screen shot 2018-05-24 at 2 52 35 pm

dkinzer:
Hiding isn't going to be easy because I'm already switching the context and merely passing the record item to the proper controller to render as it sees fit.  I'll try to fix that.

The per page thing should be an easy fix 🤞

bibliotechy:
@dkinzer Regarding the broken availability, I'm sure @sensei100 could help as she has done the most work on that on the standard results pages.

dkinzer:
@bibliotechy actually I fixed that (not pushed up yet) ... the real issue (as it turns out) is going to be pagination and sorting .  I thought bookmark would take care of that by handling it via the bookmark model.  But what is actually going on is that it passes it to the controller and let's either solr or primo handle it respectively.

Sadly that wont work for this mixed document type scenario.

If we want to keep everything in one page the easiest option is to turn pagination off and just all bookmarks appear on one page.

The second easiest thing (at this point) is to have two separate bookmark pages like we do for advanced search.

There are more options but they are not as sane.

dkinzer:
We are now getting around the API limitations by splitting the Primo ids query into sets of 9 and returning a stub document for items not found.

bibliotechy:
@dkinzer I just want to say how awesome this work is. Looks great!!! Can you please follow up by creating JIRA(s) for the outstanding work? Maybe one for pagination, if we want it, and one for the ajaxy stuff that is left to do.

dkinzer:
Thanks! will do.

PR #390, opened 2018-05-30
Fixes Safari bug in basic search form.

dkinzer:
REF BL-337
An issue with the autocomplete feature is causing a bug to surface in
Safari that disables the input field unless another field is clicked.

This commit updates the BL version to point to a branch that has the
fix.

Another possible solution is to simply disable autocomplete for this
field (I noticed that autocomplete doesn't seem to be working locally or
on the staged server, so we might want to tackle that next).

see also: projectblacklight/blacklight#1883


bibliotechy:
So, I'd prefer if the fork we were using was in the @tulibraries org.

dkinzer:
sure thing

PR #391, opened 2018-05-31
Bl 454 bento styling

sensei100:
BL-454 Updates the styling for Bento Search results
- Update fonts, borders, and background styling/colors to match mockup
- Add resource type icons to column headers
- Restrict "Books", "Articles", and "Journals" to three search results each
- Clean up formatting for search results (remove "Published:" and extra punctation)
- Move "View all" links to bottom of columns and relabel as indicated in mockup


app/search_engines/bento_search/books_engine.rb
```diff
@@ -4,7 +4,7 @@ module BentoSearch
   class BooksEngine < BlacklightEngine
     def search_implementation(args)
       query = args.fetch(:query, "")
-      query = { q: query, f: { format: ["Book"] } }
+      query = { q: query, per_page: 3, f: { format: ["Book"] } }
```
dkinzer: Instead of hard_coding the per_page here, use the per_page item passed in from args.

app/controllers/search_controller.rb
```diff
@@ -8,7 +8,7 @@ class SearchController < CatalogController
     config.add_facet_field "format", label: "Resource Type", url_method: :path_for_more_facet
   end
 
-  @@per_page = 10
+  @@per_page = 3
```
dkinzer: `@@per_page` is only getting used here and it's being injected into the individual search_engines via a param, I don't think it needs to be a global variable.

PR #392, opened 2018-06-01
Adds yarn to the docker app container.

dkinzer:


.docker/app/Dockerfile
```diff
@@ -1,10 +1,15 @@
 FROM ruby:2.4.1
-RUN apt-get update -qq && apt-get install -y build-essential libpq-dev nodejs
+RUN \
+      wget -qO- https://deb.nodesource.com/setup_9.x | bash - && \
+      wget -qO- https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - && \
+      echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list && \
+      apt-get update -qq && \
+      apt-get install -y --force-yes --no-install-recommends \
+      nodejs yarn build-essential libpq-dev nodejs
```
bibliotechy: nodejs in here twice?

dkinzer: noticed that... i've decided to replace that with jemalloc to get rid memory lock issue... i'm currently testing the idea before pushing it up.

PR #393, opened 2018-06-05
Add some items to more box.

dkinzer:
REF BL-456

Previously we were only adding the resource type facet to the More bento
box with listings for `Resource Type = Book` or `Resource Type
= Periodical/Journal` removed.

This works fine until we want to add some links to go along with the
facet.  Thus we need to generate a Sorl Query that returns results with
`Resource Type != Book` and `Resource Type != Journal/Periodical`.

In this commit, I've updated the code to append that exclusion query to
the more bento search engine query.  And, now we add the facet fields as
the last item in the results (thus it gets rendered last).

Unfortunately this query still seems to return items with `Resource Type
= Books`, even though it should not. This seems to be a bug with the
search so we need to follow up and fix that bug. See REF BL-468.

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -383,6 +383,8 @@ class CatalogController < ApplicationController
       }
     end
 
+    config.add_search_field "format", label: "Format"
+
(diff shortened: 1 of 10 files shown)


bibliotechy:
Looks great!

PR #395, opened 2018-06-05
BL-454 Updates to bento styles

sensei100:
BL-454 Updates styles for bento results
- Change font for headers
- Add bold and underlines to total_results
- Update journal icon


dkinzer:
This could be an issue on my local, but I'm not seeing the online button anymore, it's a link and it's shifted to the left.

image

sensei100:
@dkinzer I think that got lost when I fixed the merge conflict.  I added the styles back in now.

PR #396, opened 2018-06-05
Tweaks relevance results per BL-447

bibliotechy:
Adds relevance tests based on examples provided.

Adds a custom matcher so we can write tests like

 `expect(result).to include_docs([1,2,3]).before([4,5])``

Bosst a bit based on recency, but not too much. 

Add boosts for less sloppy TOC and author.

Also closes BL-430.

Moves relevance tests into separate test suite to be optionally run.


dkinzer:
This looks great.  There are some merge conflicts, though.

dkinzer:
This looks great.  I wont wait for tests to pass again since the last update is just to a README file and the tests have already passed.

PR #399, opened 2018-06-06
Try to fix libqa routing issue.

dkinzer:
REF BL-414

Adding a similar configuration to the app router fixed a similar issue
with primo_central tracking so I'm hopping this will do the same for
catalog/bookmarks tracking on libqa.

For example the following track requests fail with 404:

```
https://libqa.library.temple.edu/catalog/catalog/991017730139703811/track?counter=1&search_id=2708

https://libqa.library.temple.edu/catalog/TN_crossref10-dot-1177-slash-0361684312464862/track?counter=4
```

We should probably add a redirect for catalog/:id/track where :id
matches /^TN_/, too.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -35,6 +35,7 @@ Rails.application.routes.draw do
     concerns :exportable
   end
 
+  post "catalog/:id/track" => "catalog#track"
   post "articles/:id/track" => "primo_central#track", as: :track_primo_central
 
   resources :users, only: [:index] do


bibliotechy:
Blacklight already includes that route https://github.com/projectblacklight/blacklight/blob/master/config/routes.rb#L5

If you run `rake routes` you can see:
```
Routes for Blacklight::Engine:
...                       
                                   POST   /saved_searches/forget/:id(.:format) saved_searches#forget
track_search_context POST   /catalog/:id/track(.:format)         catalog#track <-----------
           suggest_index  GET    /suggest(.:format)                   suggest#index {:format=>"json"}
```

dkinzer:
@bibliotechy hmm... it's getting a 404 on those URLs .. 

Maybe the difference is that those paths have a root "/catalog/:id/track" and this one is relative? ("catalog/:id/track")

PR #400, opened 2018-06-06
BL-459 Updates navbars to include catalog/articles

sensei100:
BL-459 Updates include:
- Adding Catalog/Articles headers to nav bars
- Changing the headers on the advanced search pages
- Adding a link to articles/catalog search
screen shot 2018-06-06 at 3 02 12 pm


dkinzer:
I like this feature.

PR #401, opened 2018-06-06
Set up rails app with url prefix.

dkinzer:
See
https://medium.com/@brain.amsterdam/how-to-run-rails-application-with-url-prefix-281e1dda37d4

This might not be the exact recommendation.


dkinzer:
This seems like the wrong way

PR #408, opened 2018-06-07
BL-469 Add physical_material_type

sensei100:
BL-469 Adds physical_material_type from the alma api to certain types of records
- It should not display for Book, Issue, Music Score, Kit, Map, Bound Issue, Government Document, Other
- Displays in the third column of the availability table
screen shot 2018-06-07 at 2 19 07 pm

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/tulibraries/alma_rb
-  revision: 6060ca25486bd1fc24f8325d6e3a233fedeafb8e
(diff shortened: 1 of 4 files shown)


dkinzer:
@sensei100 this isn't showing up for me locally, is there a specific item I need to look at?

sensei100:
@dkinzer It should be on any dvd.

sensei100:
@dkinzer You may need to bundle update alma to get the recent changes there too.

dkinzer:
I did bundle install.. but I'll try bundle update instead.

dkinzer:
All sorted out now.  Looks good!

PR #409, opened 2018-06-07
Update more beto box.

dkinzer:
REF BL-474

Followup changes to bento more box issue.
* Limit results to 2.
* Add view all catalogs results link.

### DIFF
--- a/app/search_engines/bento_search/more_engine.rb
+++ b/app/search_engines/bento_search/more_engine.rb
@@ -4,9 +4,7 @@ module BentoSearch
   class MoreEngine < BlacklightEngine
     def search_implementation(args)
       query = args.fetch(:query, "")
-      per_page = args.fetch(:per_page)
-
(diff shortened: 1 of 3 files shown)


app/search_engines/bento_search/more_engine.rb
```diff
@@ -4,9 +4,7 @@ module BentoSearch
   class MoreEngine < BlacklightEngine
     def search_implementation(args)
       query = args.fetch(:query, "")
-      per_page = args.fetch(:per_page)
-
-      query = { q: query, per_page: per_page, seach_field: "advanced" }
+      query = { q: query, per_page: 2, seach_field: "advanced" }
```
sensei100: @dkinzer It looks like there is a typo here.  Should it be search_field?

dkinzer: Ah, that explains some things.  Great catch.  We can actually remove that altogether.

PR #412, opened 2018-06-08
BL-469 Display desc, not value

sensei100:
BL-469 Physical material type should use desc rather than value
- 991025183159703811 should display as Sound Recording, not RECORD
screen shot 2018-06-08 at 8 53 02 am

### DIFF
--- a/app/helpers/alma_data_helper.rb
+++ b/app/helpers/alma_data_helper.rb
@@ -34,8 +34,8 @@ module AlmaDataHelper
   end
 
   def physical_material_type(item)
-    type = item.physical_material_type["value"]
-    if item.physical_material_type.present? && !type.match(PHYSICAL_TYPE_EXCLUSIONS)
+    type = "#{item.physical_material_type["value"]}"
+    if !type.match(PHYSICAL_TYPE_EXCLUSIONS)
       return item.physical_material_type["desc"]
     end
   end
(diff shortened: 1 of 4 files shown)


config/environments/development.rb
```diff
@@ -13,10 +13,10 @@
   # In the development environment your application's code is reloaded on
   # every request. This slows down response time but is perfect for development
   # since you don't have to restart the web server when you make code changes.
-  config.cache_classes = true
+  config.cache_classes = false
```
dkinzer: I think this got in by mistake.

app/helpers/alma_data_helper.rb
```diff
@@ -34,8 +34,8 @@ def description(item)
   end
 
   def physical_material_type(item)
-    if item.physical_material_type.present? && !PHYSICAL_TYPE_EXCLUSIONS.match(item.physical_material_type)
-      return item.physical_material_type
+    if item.physical_material_type.present? && !PHYSICAL_TYPE_EXCLUSIONS.match(item.physical_material_type["value"])
```
dkinzer: I believe this bit is can be more succinctly expressed

`if !"#{item&.physical_material_type["value"]}".match(PHYSICAL_TYPE_EXCLUSION)`

Or for readabilities sake

```
type = "#{item&.physical_material_type["value"]}"
if !type.match(PHYSICAL_TYPE_EXCLUSIONS)
```

sensei100:
@dkinzer I included the wrong file in my last commit.  Sorry about that.  I reverted the config file and added the spec file that I intended to commit.  I also made the changes to the method that you requested.

PR #414, opened 2018-06-08
BL-474 Update styles for bento

sensei100:
BL-474 Changes to bento results styling:
- Remove gray background from Resource types and make text smaller
- Add width to columns to align with basic search input and my library account


dkinzer:
😎 I like this looks much better!

PR #418, opened 2018-06-11
Fix tests that break when relevance tests are run.

bibliotechy:
Made a few tests less brittle to more records being present.

Removed a few duplicated records in the test_search fixture


dkinzer:
This seems like it's still error-ing out in travis.

PR #419, opened 2018-06-11
Update traject mappings

sensei100:
BL-475 Adds subfield I to contributor_display and title_addl_display
- Requires a reindex


dkinzer:
I've restarted travis build, because I think the build just got stuck for some reason.

PR #421, opened 2018-06-11
Add checkbox for easy "Online" resource search.

dkinzer:
REF BL-460

Adds a checbox that when clicked toggles the faceted Online search for
both the Catalog and the Articles searches.


bibliotechy:
As per comments on BL-460, this check box should only be present for SOlr search, not Articles search

PR #428, opened 2018-06-13
Advanced search

cdoyle-temple:
#BL-479
Tweak layout for boolean operators in Advanced Search (Catalog)


cdoyle-temple:
Should be good now, David.

PR #432, opened 2018-06-13
Change pub date from field 260 to CF 008/007-010

skng5:
Issue #BL-205

Note:
Issues refers to date type and pub status. Code in place to handle future uses for that field.


dkinzer:
@skng5 2 tests are failing in travis for this (we still haven't fixed whatever bug is causing failures not to error out on travis so green on github is misleading)

PR #434, opened 2018-06-14
Index record availability and skip unavailable items.

dkinzer:
REF BL-480
* Updates the traject indexer configuration to capture a record's
availability.
* Skips building a link to unavailable records.
* Removes premature compact of data that breaks structural integrity of
the electronic_resource_display field.

### DIFF
--- a/lib/traject/macros/custom.rb
+++ b/lib/traject/macros/custom.rb
@@ -148,7 +148,7 @@ module Traject
       def extract_electronic_resource
         lambda do |rec, acc, context|
           rec.fields("PRT").each do |f|
-            selected_subfields = [f["a"], f["c"], f["g"]].compact.join("|")
+            selected_subfields = [f["a"], f["c"], f["g"], f["9"]].join("|")
             acc << selected_subfields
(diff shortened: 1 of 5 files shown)


lib/traject/macros/custom.rb
```diff
@@ -148,7 +148,7 @@ def extract_subject_topic_facet
       def extract_electronic_resource
         lambda do |rec, acc, context|
           rec.fields("PRT").each do |f|
-            selected_subfields = [f["a"], f["c"], f["g"]].compact.join("|")
+            selected_subfields = [f["a"], f["c"], f["g"], f["9"]].join("|")
```
bibliotechy: Is the compact that was removed no longer necessary?

dkinzer: We should not be compacting at this stage because we make assumption late about the structure of splitting on pipe and compacting here potentially messes with that structure.

bibliotechy: 👍 
Just being cautious.

PR #442, opened 2018-06-14
Bl 388 solr timestamp based updates

relaxing:
Turns on Solr doc-centric versioning on date, and populates the record update date from the Alma ADM b field.


relaxing:
Darn. Well no thanks to you Travis. 
 @dkinzer @bibliotechy Is there a way to make Traject send a default for a field if none is specified? 
Either that or we regenerate our test data from the newer Alma exports.
I tried to give the Solr schema a default for the timestamp, but that didn't satisfy it.

dkinzer:
@relaxing I posted your question on the #traject code4lib slack channel and got this response from the maintainer of traject:

```
# You have to make sure this is at the END of your traject pipeline
each_record do |record, context|
  if context.output_hash["some_field"].nil? || context.output_hash["some_field"] == []
    context.output_hash["some_field"] = ["default_value"]
  end
end
```

PR #443, opened 2018-06-14
Flash notice

cdoyle-temple:
Resetting flash login notices to previous behavior, changes did not work as expected.


dkinzer:
closing as #444 was the real change.

PR #447, opened 2018-06-15
Change successful login message flash to success.

cdoyle-temple:
This should work. I found a way to do it locally and it looked good. *but* if it doesn't work here I will revert this change.


dkinzer:
OK, I tried this locally and it works fine.  Great job! 

Demise really needs to update its docs.

PR #452, opened 2018-06-15
Do not exit with 0 when tests fail.

dkinzer:
Fixes tests failing on travis but passing build.

### DIFF
--- a/lib/tasks/server.rake
+++ b/lib/tasks/server.rake
@@ -16,11 +16,12 @@ task :rspec, [:spec_args] do |t, args|
     run_solr("test", port: "8985") do
       Rake::Task["fortytu:solr:load_fixtures"].invoke if ENV["DO_INGEST"]
       rspec_cmd = ["rspec", args[:spec_args]].compact.join(" ")
-      system(rspec_cmd)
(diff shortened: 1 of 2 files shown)


lib/tasks/server.rake
```diff
@@ -12,15 +12,17 @@ end
...
   if Rails.env.test?
     run_solr("test", port: "8985") do
       Rake::Task["fortytu:solr:load_fixtures"].invoke if ENV["DO_INGEST"]
       rspec_cmd = ["rspec", args[:spec_args]].compact.join(" ")
-      system(rspec_cmd)
+      passed = system(rspec_cmd)
     end
   else
-    system("rake rspec[#{args[:spec_args]}] RAILS_ENV=test")
+    passed = system("rake rspec[#{args[:spec_args]}] RAILS_ENV=test")
```
dkinzer: I think the culprit in our woes is this `system("rake rspec` which will eventually call `system` again.  It turns out `system` does not actually result in `exit 1` status .. it just returns true or false.  So I'm guessing that the bit of inception here is causing rspec to miss that `system` is returning false in one of these cases.

PR #455, opened 2018-06-15
Availability facet

cdoyle-temple:
#BL-487 Close "Availability" by default in Catalog search


cdoyle-temple:
#BL-487 Close "Availability" by default in Catalog search

PR #477, opened 2018-06-20
Added fixed widths to availability table cells.

cdoyle-temple:
Priority to location, callnum, desc, and availability. Type and Public note are tiny.


cdoyle-temple:
Also made tucob.css into sass stylesheet (.scss)

PR #487, opened 2018-06-22
Remove dashes from names in subject display.

dkinzer:
REF BL-351

Require a re-index.


bibliotechy:
Seeing dashes after single part subjects 
image
Example item I looked at was 991012169639703811, which is int the spec fixtures

dkinzer:
Ooh, thanks for adding the specific example.

dkinzer:
@bibliotechy updated to take care of issue when items are single.

PR #491, opened 2018-06-26
Allow two minutes to read network files.

dkinzer:
Update the read timeout for harvest task to 2 minutes.

The harvest task uses HTTParty which in turn uses the Net::HTTP default
of 60 seconds.

This commit doubles the default read timeout to 120 seconds.

Also, this commit forces an exit if harvest throws an error which should
fix the false positive we see on network errors.


bibliotechy:
I cherry picked this commit for https://github.com/tulibraries/tul_cob/releases/tag/v0.2.2, so I'm going to rebase master on top of it to keep releases and master aligned

PR #492, opened 2018-06-26
Add ajax lookup of failed Primo bookmark.

dkinzer:
REF BL-457


bibliotechy:
It is loading the the expected bookmark, but the numbering is off
image

dkinzer:
@bibliotechy, I updated the code to render the correct item count.

PR #494, opened 2018-06-28
Handle record not found with 404.

dkinzer:
REF BL-499


bibliotechy:
@dkinzer So, we actually explicitly handle 404's and 500s in the errors controller so that we get a dynamic error pages to maintain the site styling.   https://github.com/tulibraries/tul_cob/commit/5f1a6990da220a6a95025eb2bc24e055b4852f68#diff-c972bc2ac8aea20490c3d38669180d96

dkinzer:
@bibliotechy I'm not sure that's working as intend the following should return 400 but instead it returns 500

```
curl -I https://librarybeta.temple.edu/catalog/failme
HTTP/1.1 500 Internal Server Error
Date: Thu, 28 Jun 2018 18:27:58 GMT
Server: Apache/2.4.6 (CentOS) OpenSSL/1.0.2k-fips Phusion_Passenger/5.3.2
Referrer-Policy: strict-origin-when-cross-origin
X-Permitted-Cross-Domain-Policies: none
X-XSS-Protection: 1; mode=block
X-Request-Id: 85791126-41f5-41b1-ad45-799e3b4e5489
X-Download-Options: noopen
X-Runtime: 0.067752
X-Frame-Options: SAMEORIGIN
X-Content-Type-Options: nosniff
X-Powered-By: Phusion Passenger 5.3.2
Strict-Transport-Security: max-age=63072000; includeSubDomains; preload
Content-Length: 8287
Status: 500 Internal Server Error
Connection: close
Content-Type: text/html; charset=utf-8
```

Also blacklight has a `rescue_from` method for missing ids that I think we may need to override or the Exception will continue to get passed to honeybadger for missing ids.

There is one other issue this is dealing with but I'll separate it out to it's own PR so that it doesn't get lost with this one.

dkinzer:
@bibliotechy I've updated this PR so that it uses our error handling code instead of blacklight's which was causing the 500 error.

Also I removed the primo central change and moved that to it's own PR.

PR #503, opened 2018-07-12
Update from email address.

dkinzer:
REF BL-241

I'm updating the from: email to asktulibrary@temple.edu.  I'm not sure
this will work because I did not update the smtp user/password config.
But I'm guessing we can allow books@temple.edu to send emails as
asktulibrary in some GMAIL config panel.


bibliotechy:
@dkinzer Waiting on @ebtoner to provide credentials before I merge this.

PR #504, opened 2018-07-13
Add sorting fields to articles.

dkinzer:
REF BL-514

Adds sorting to article searches.  The fields used are the same as what
is currently available in the catalog search.

I've implemented this as per the (limited) documentation on the API. As
far as I can tell it does not seem to work as I would expect it.  I'm
hoping there is a configuration change we can make on the API itself to
make the sorting better.

I read the following documentation (not pnxs specific) that leads me to
believe sorting can only happen with the full view API so I'm making
sure to make that the default for our search:

https://knowledge.exlibrisgroup.com/Primo/Knowledge_Articles/Default_sorting_in_Primo_Brief_Result_List

Never the less, I don't think it made much of a difference.

This may be an API bug we will have to address with ExLibris.


bibliotechy:
@dkinzer I'm not seeing a difference either. I believe this is a bug with the Primo API. I'm going to file it and issue. I'm think we should close this PR but preserve the branch so we can use it again if required.

PR #516, opened 2018-07-18
Redirect user to "users/account" after login.

dkinzer:
REF BL-522

This commit adds a way to generate login forms that can dynamically
redirect the user to a specified path.

The current default redirect path for the My Accounts link has been set to
"/users/account".

### DIFF
--- a/app/views/shared/_nav_bar_login_links.html.erb
+++ b/app/views/shared/_nav_bar_login_links.html.erb
@@ -4,6 +4,6 @@
     </li>
 <% else %>
     <li>
-    <%= link_to t('blacklight.header_links.login'), new_user_with_redirect_path("/users/account") %>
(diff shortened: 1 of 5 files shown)


app/views/shared/_nav_bar_login_links.html.erb
```diff
@@ -4,6 +4,6 @@
     </li>
 <% else %>
     <li>
-      <%= link_to t('blacklight.header_links.login'), new_user_session_path %>
+    <%= link_to t('blacklight.header_links.login'), new_user_with_redirect_path("/users/account") %>
```
bibliotechy: Using hard coded paths causes problems when we deploy to subdirectories. Is there a reason to not use `new_user_session_path` as the parameter for the new method?

dkinzer: Thanks, I meant to use that just forgot to.

PR #517, opened 2018-07-19
Bl 382 hold requests

sensei100:
BL-382 Hold request functionality 
- Both bib level and item level requests are sending
- Updated show controller for modals on the full record page


sensei100:
@bibliotechy I adjusted the columns, so it should be okay now.

PR #518, opened 2018-07-20
Bl 516 explanatory text

sensei100:
BL-516 Adds text for users to see what is included in their search results

### DIFF
--- a/app/assets/stylesheets/tucob.scss.erb
+++ b/app/assets/stylesheets/tucob.scss.erb
@@ -27,6 +27,16 @@ h1.advanced.page-header a {
   display:none;
 }
 
+.btn-link {
+  color: #A50030;
+}
+
+.btn-link:hover {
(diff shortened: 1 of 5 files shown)


bibliotechy:
In order for the messages to remain hidden, we need to set and check a cookie before displaying these. This can be done with the `session` object available in the controller.

So, when someone clicks Hide, the stimulus controller should hide the text and then POST a request to a rails controller that sets the session variable for that page to hidden. Something like

```
def hide_search_explanation
page = params[:page] #a param we'll pass about what page we are on
cookie_name = "#{page}_search_explain".to_sym
session[cookie_name] = "hide"
end
```
then in the templates where the message displays, add some logic
```
if session[cookie_name] == "hide"
  search explainer goes here
end
```

sensei100:
@bibliotechy I created different methods for each view, so I think this one is ready for review again.

PR #532, opened 2018-07-31
Update alma gem.

dkinzer:
REF BL-518


dkinzer:
This can be merged now.

PR #533, opened 2018-07-31
Fix ampersand/and result divergence with stopwords

bibliotechy:
Add qf and pf for title_statement_* search fields.

  Tweak mm lower for larger queries
  Please enter the commit message for your changes. Lines starting


dkinzer:
Locally I'm getting the same result set but in different order for "date AND test" vs. "date & test".  But from a purely code review perspective this LGTM, so I'm merging and leaving it up to @ebtoner to decide if further work is required after re-indexing is done.

PR #539, opened 2018-08-02
Add login for visitors

bibliotechy:

PR #540, opened 2018-08-02
Redirect back to referrer.

dkinzer:
REF BL-550


bibliotechy:
Superseded by #539 which was rebased on top of this

PR #550, opened 2018-08-07
Bl 544 styling updates

sensei100:
Updates the placement of the request buttons on search result and record show pages

### DIFF
--- a/app/assets/stylesheets/tucob.scss.erb
+++ b/app/assets/stylesheets/tucob.scss.erb
@@ -161,14 +161,27 @@ input[type="radio"] + label {
   display: none;
 }
 
-.availability-panel-heading {
-  background-color: #A30733 !important;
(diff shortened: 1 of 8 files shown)


dkinzer:
@sensei100 this change is looking looking different for articles than it does for catalog ...
image

sensei100:
@dkinzer I need to remove the panel-default class from  <div id="record-view-iframe" class="panel panel-default">, but I can't figure out where it is coming from.  Can you point me to where that would be?

dkinzer:
@sensei100 OK, I'm seeing that in two places:

```
app/views//catalog/_show_availability_section.html.erb:4:<div id="record-view-iframe"  class="panel panel-default">
app/views//primo_central/_show_availability_section.html.erb:3:<div id="record-view-iframe"  class="panel panel-default">
```

PR #554, opened 2018-08-07
Bl 506 item requests

sensei100:
BL-506 Remove item level requests for booking
- add logic to handle errors when someone else has already booked an item

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -13,7 +13,7 @@ Rails.application.configure do
   # In the development environment your application's code is reloaded on
   # every request. This slows down response time but is perfect for development
   # since you don't have to restart the web server when you make code changes.
-  config.cache_classes = false
+  ENV["CACHE_CLASSES"] ? true : false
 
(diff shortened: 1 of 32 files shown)


config/environments/development.rb
```diff
@@ -13,10 +13,10 @@
   # In the development environment your application's code is reloaded on
   # every request. This slows down response time but is perfect for development
   # since you don't have to restart the web server when you make code changes.
-  config.cache_classes = true
+  config.cache_classes = false
```
dkinzer: I'm wondering if we should just keep this as is and only change when we are working on bento specifically.  What do you think @bibliotechy ?

sensei100: @dkinzer Sorry, that file isn't supposed to be there.  I will take it out.

bibliotechy: Maybe set it to `ENV["CACHE_CLASSES"] ? true : false` so we can override it locally if need be, if we are working on Bento stuff, but normally it is turned off.

PR #567, opened 2018-08-08
Analytics

cdoyle-temple:
#557 Event listeners only added if element exists on page being viewed.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -151,3 +151,13 @@ function toggleOnlineOnly() {
     online_facet.click()
   }
 }
+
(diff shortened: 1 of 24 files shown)


cdoyle-temple:
But they do not all have the same category: i.e. some will have "navigation", some "catalog" or "articles", etc, as we try to organize Google Analytics in future.

dkinzer:
@cdoyle-temple  That was just an example for the simplest case.  It would still be possible to pull out all the ids with relevant info to a data structure that you can iterate on.

```javascript
var items = [ { id: "id1", stuff: "navigation", .. }, ...]
```

PR #576, opened 2018-08-09
Fix bug with pickup locations

sensei100:
Pickup locations were being limited to only owning library.  
- equipment should be limited to owning library
- all other resource types should default to valid pickup locations

### DIFF
--- a/app/lib/cob_alma/requests.rb
+++ b/app/lib/cob_alma/requests.rb
@@ -63,12 +63,15 @@ module CobAlma
     end
 
     def self.equipment(items_list)
+      pickup_locations = []
       items_list.each do |item|
         if item.circulation_policy == "Equipment"
-          pickup_locations = [item.library, item.library_name]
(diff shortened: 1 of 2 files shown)


app/lib/cob_alma/requests.rb
```diff
@@ -63,12 +63,15 @@ def self.valid_pickup_locations(items_list)
     end
 
     def self.equipment(items_list)
+      pickup_locations = []
       items_list.each do |item|
         if item.circulation_policy == "Equipment"
-          pickup_locations = [item.library, item.library_name]
+          pickup_locations << item.item_data.fetch("library")
+        else
+          pickup_locations = []
```
dkinzer:
```
else
  pickup_locations = []
```
is not necessary since you are already instantiating it on line 66.

sensei100: @dkinzer Good catch.  I took it out.

PR #580, opened 2018-08-09
Bl-317 allows boundwith records to have availability shown and to be requested 

bibliotechy:
This PR allows :wastebasket: :fire:  records, otherwise known as boundwith records, to have availability information displayed and be requested.

Did I mention they were a total :wastebasket: :fire: ?

It also adds logic to skip Host (parent) records so they are not skipped.

Whoever takes this on to review, I can send you the marc file with the required records that are created in a separate process

Just for good measure
:wastebasket: :fire: 
:wastebasket: :fire: 
:wastebasket: :fire: 
:wastebasket: :fire: 

Also a :garbagefire:

### DIFF
--- a/spec/fixtures/marc_fixture.xml
+++ b/spec/fixtures/marc_fixture.xml
@@ -28599,6 +28599,7 @@
     </datafield>
   </record>
   <record>
+<<<<<<< 2d5d2803ac7ba908918d908356fcd3d287af68bc
     <leader>05101cpcaa2200709Ia 4500</leader>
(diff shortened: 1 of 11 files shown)


spec/fixtures/marc_fixture.xml
```diff
@@ -28599,6 +28599,7 @@
     </datafield>
   </record>
   <record>
+<<<<<<< 2d5d2803ac7ba908918d908356fcd3d287af68bc
```
dkinzer: Still seeing a merge conflict here.

PR #603, opened 2018-08-15
Add specs for requests

sensei100:
Adds specs to test for exceptions
- also refactors dates to rescue empty strings


bibliotechy:
@sensei100 let's walk through a quick refactor together after lunch.

PR #606, opened 2018-08-16
Bl 580 truncate titles

sensei100:
Shortens titles in the search results to 300 characters
- see 991004250539703811 in the marc_fixture.xml file for an example
screen shot 2018-08-16 at 9 05 20 am

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -14,6 +14,11 @@ module ApplicationHelper
     render_location(value[:value].first)
   end
 
+  def truncated_link(document, field, opts = { counter: nil })
+    label = index_presenter(document).label(field, opts).truncate(300).html_safe
+    link_to label, url_for_document(document), document_link_params(document, opts)
+  end
+
   def get_search_params(field, query)
     case field
     when "subject_display"
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -15,7 +15,7 @@ module ApplicationHelper
   end
 
   def truncated_link(document, field, opts = { counter: nil })
-    label = index_presenter(document).label(field, opts).truncate(300).html_safe
+    label = index_presenter(document).label(field, opts).truncate(300, separator: " ").html_safe
     link_to label, url_for_document(document), document_link_params(document, opts)
   end
 
--- a/app/views/catalog/_index_header_default.html.erb
+++ b/app/views/catalog/_index_header_default.html.erb
@@ -1,0 +1,22 @@
+<%# header bar for doc items in index view -%>
+<header class="documentHeader row">
+  <%# main title container for doc partial view
+      How many bootstrap columns need to be reserved
+      for bookmarks control depends on size.
+  -%>
+  <% document_actions = capture do %>
+    <% # bookmark functions for items/docs -%>
+    <%= render_index_doc_actions document, wrapping_class: "index-document-functions col-sm-3 col-lg-2" %>
+  <% end %>
+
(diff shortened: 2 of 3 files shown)


app/helpers/application_helper.rb
```diff
@@ -14,6 +14,11 @@ def render_location_show(value)
     render_location(value[:value].first)
   end
 
+  def truncated_link(document, field, opts = { counter: nil })
```
dkinzer: I think it makes more sense to override the link_to_document helper method in blacklight instead of creating a new helper method.  That way we don't also have to override the blacklight template.

sensei100: @dkinzer I see what you are saying, but my concern is that this method does seem to be used in multiple places.  The only place that we wanted truncation was in the search results title.  I also like the fact that the method_name clearly lets someone looking at the code what is happening there.  What are your thoughts about that?

dkinzer: @sensei100 sure thing, here are my thoughts:
* link_to_document only get's use in two places .. the template for creating document link on index results page and for creating document_link in the _show_more_like_this template.  I'm not sure we are even using the _show_more_like_this template but even if we are I think we will need to treat the links to documents in the same way thus necessitating another template override.
* Regarding the method name I think `link_to_document` is more descriptive than `truncated_link` in this case; I suppose we could use `trunkcated_link_to_document` but I still think we should just override the `link_to_document`

I would like to hear @ebtoner's opinion on this since her comment from this morning suggested that we might want to handle this at indexing.

sensei100: @dkinzer The ticket says that it should be done at the Blacklight level, not SOLR.

dkinzer: I can verify it is not getting used for thumbnails.

dkinzer: @sensei100 @ebtoner I do actually like that handling this on BL does take care of any similar issues in Primo article titles as well.

app/helpers/application_helper.rb
```diff
@@ -14,6 +14,11 @@ def render_location_show(value)
     render_location(value[:value].first)
   end
 
+  def link_to_document(document, field, opts = { counter: nil })
+    label = index_presenter(document).label(field, opts).truncate(300, separator: " ").html_safe
```
dkinzer: @sensei100 I think we should duplicate more of how `link_to_document` is currently implemented, or there could be unforeseen issues.  The BL implementation updates the value of `field` in some cases:
https://github.com/projectblacklight/blacklight/blob/v6.14.1/app/helpers/blacklight/url_helper_behavior.rb#L20

The blacklight version has

sensei100: @dkinzer This is the reason why I feel the original approach of having a different method is preferable.  There are no unforeseen consequences and the method will only be used where we want it to.

dkinzer: @sensei100 yeah, I hear you; and that makes total sense. I guess part of what I was thinking was that we could just copy the function as is and then just make that minor change.  Part of me thinks it could be a nice contribution back to BL because I doubt anyone would want title links to document that are very long.

But the more I think about it the more I wonder that they will just say something like "we already let you modify that using field processor via config.

Let me mull this over for a little bit.  I want to see if it's possible to truncate via configuration.  We could do a hangout if you want to look into it together?

dkinzer: @sensei100 OK, what do you think about this as an alternative? https://github.com/tulibraries/tul_cob/commit/953836a02077

dkinzer: @sensei100 what's the status on this?

PR #610, opened 2018-08-16
Add a back_to_search button on bookmarks page

sensei100:
BL-581 Links back to search from bookmarks page

### DIFF
--- a/app/assets/stylesheets/tucob.scss.erb
+++ b/app/assets/stylesheets/tucob.scss.erb
@@ -452,6 +452,14 @@ ul.navbar-nav {
   display:none!important;
 }
 
+.bookmarks-back-btn {
+  display: inline-block;
+  margin-left: 15px;
+  background-color: #f1efe7 !important;
+  color: #A50030;
+  padding: 8px 12px;
+}
+
 /* === MEDIA QUERIES === */
 
 @media (min-width: 992px) {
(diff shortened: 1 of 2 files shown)


ebtoner:
@dkinzer What happens when a user goes directly to https://library.temple.edu/library-search-faq/beta-feedback? Would the "Back to Search" button be omitted like it is for that button on the full record page? If so, I'm inclined to leave it as "Back to Search" for now.

dkinzer:
@ebtoner sorry, I'm not following, this "Back to Search" button is specific to the bookmarks page.

ebtoner:
We're replicating this button from the one used on the full record pages, though. I'm wondering if the one in Bookmarks will (or could) have the same conditional display.

sensei100:
@ebtoner @dkinzer I was unable to use the same method that the link on the full record pages uses.  It changed the text of the button to Back to Bookmarks and would keep you on the bookmark page when you click it.  So I can't guarantee that it would have the exact same behavior.  I'm not sure how someone would go directly to the bookmarks from the page that you provided though.

ebtoner:
Ooops --- now I see why you both are confused! I provided the wrong link (clearly I'm ready for the weekend). I meant to use https://librarysearch.temple.edu/bookmarks, i.e. when someone might directly access the URL (perhaps if they had saved it in their browser).

ebtoner:
In short, though, let's go ahead and use "Back" then, if that conditional display doesn't work.

sensei100:
This is what they would see: 
screen shot 2018-08-17 at 10 37 13 am

sensei100:
Sorry, that screen shot was taken on the wrong branch.  You would still see the link if you went directly there.

dkinzer:
I think as long as there are bookmarks you will see the link.  But in my opinion that should be OK.

ebtoner:
Agreed.

PR #618, opened 2018-08-20
Add validation to empty advanced form

sensei100:
Prevents the user from submitting an empty form in advanced article search


dkinzer:
This works; note however that locally if I click on a field I get javascript error that then breaks this functionality but I think it's not related other than that if javascript is not working then this will fail.

PR #622, opened 2018-08-21
Blacklight request modal

sensei100:
Switch over to using the blacklight modals for requests


dkinzer:
@sensei100 when I click on availability button I'm getting ```target_properties.js:13 Uncaught (in promise) Error: Missing target element "availability.href"
    at Controller.get (target_properties.js:13)
    at availability_controller.js:31``` error in the console.

Also I'm seeing a bunch of errors that look like
```
{error: {…}}
error
:
{errorCode: "402203", errorMessage: "Input parameters mmsId 991036740750603811 is not valid.", trackingId: "E02-2408150039-QDBLR-AWAE1497149572"}
__proto__
:
Object
```
in the console.  I can't reproduce this on librarysearch.temple.edu so I'm assuming they are part of this PR.

dkinzer:
OK, let me try again could be a local thing.

dkinzer:
OK, I'm still seeing it.  Here is what I am doing to reproduce it.

1. Go to search page
2. Search for "primes"
3. Click on the first "Available" result.

sensei100:
@dkinzer I did that same search and do not see any console errors.  I did get an error when trying to request it, but that is an unrelated bug that I discovered today.  That book is at Paley and Ambler.  Ambler should not be showing up as a pickup location, but it does.  If I try to use Ambler I get an error, but if I select another library the request goes through.  I'll make a JIRA ticket for the other issue, but that isn't due to something in this PR.  I'm not sure why you are seeing a console error though.

dkinzer:
@bibliotechy thanks for checking.  The issue must be with my local.

PR #626, opened 2018-08-23
Update primo_central_document to use new api.

dkinzer:
REF BL-606


bibliotechy:
@dkinzer Ready for the update reference to primo gem

dkinzer:
thanks! will do.

PR #629, opened 2018-08-24
Bl 609 aeon requests

sensei100:
This moves the aeon button back into the availability panel
- Button should display only for SCRC materials and will be on the same row as the library name
screen shot 2018-08-24 at 2 00 13 pm


sensei100:
@bibliotechy We do need the default. Sorry, I missed that when switching things to the old version.

PR #630, opened 2018-08-24
Add sorting fields to articles.

dkinzer:
REF BL-514

Adds sorting to article searches.  The fields used are the same as what
is currently available in the catalog search.

Neither `creator` nor `date2` sorting is working.  We might need to
bring it up with Exlibris.

Depends on #626


bibliotechy:
@dkinzer I'm not seeing the sort drop downs have any effect. Sort order remains the same no matter what sort option I choose.

dkinzer:
Hmmm... could be it be caching?

bibliotechy:
That's a possibility.

PR #631, opened 2018-08-27
Only show date format message for unsupported browsers

sensei100:
BL-608 The format date message could be confusing for users that see the date presented in a different format
- limit the message to only browsers that don't support the date field (safari and internet explorer)
-safari view 
screen shot 2018-08-27 at 9 49 11 am
- chrome view 
screen shot 2018-08-27 at 9 50 05 am


sensei100:
@bibliotechy That makes sense.  I refactored it into a partial.

PR #633, opened 2018-08-28
Show availButton for lost or missing items

sensei100:
The availability button was never leaving the loading state for items with empty holdings
- You should now see the "Not Available" collapse button for those records
screen shot 2018-08-28 at 2 30 47 pm

### DIFF
--- a/app/assets/javascripts/blacklight_alma/blacklight_alma_lib.js
+++ b/app/assets/javascripts/blacklight_alma/blacklight_alma_lib.js
@@ -42,6 +42,15 @@ var BlacklightAlma = function (options) {
    }
  }
 
+ noHoldingsAvailabilityButton = function(id) {
+   var availButton = $("button[data-availability-ids='" + id + "']");
+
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/blacklight_alma/blacklight_alma_lib.js
```diff
@@ -42,6 +42,15 @@ var BlacklightAlma = function (options) {
    }
  }
 
+ noHoldingsAvailabilityButton = function(id) {
+   var availButton = $("button[data-availability-ids='" + id + "']");
+
+   $(availButton).html("<span class='avail-label not-available'>Not Available</span>");
```
bibliotechy: Looks like these 4 lines duplicate https://github.com/tulibraries/tul_cob/blob/60d2806c9d27c448d474782123968d8d02fedf8e/app/assets/javascripts/blacklight_alma/blacklight_alma_lib.js#L37-L40. Could they be refactored into a function that is called in both places?

sensei100:
@bibliotechy I refactored to make that change.

PR #637, opened 2018-08-30
Add polyfill function to have javascript includes in ie 11

sensei100:
BL-585 Stimulus polyfill package doesn't handle the includes method
- add function to blacklight_alma to take care of include

### DIFF
--- a/app/assets/javascripts/blacklight_alma/blacklight_alma_lib.js
+++ b/app/assets/javascripts/blacklight_alma/blacklight_alma_lib.js
@@ -370,3 +370,14 @@ var BlacklightAlma = function (options) {
          $(element).html("<span style='color: red'>No status available for this item</span>");
      });
  };
+
+ /** This is a polyfill to get rid of a console error in Internet Explorer 11
+ that is preventing the availability information from being shown
+ */
+ String.prototype.includes = function(value){
+   return this.indexOf(value)>-1
+ }
+
+ Array.prototype.includes = function(value){
+   return this.indexOf(value)>-1||!1
+ }


sensei100:
@dkinzer That makes sense. I added conditional logic to limit it's use.

app/assets/javascripts/blacklight_alma/blacklight_alma_lib.js
```diff
@@ -370,3 +370,18 @@ var BlacklightAlma = function (options) {
...
+ /** This is a polyfill to get rid of a console error in Internet Explorer 11
+ that is preventing the availability information from being shown
+ */
+if (!String.prototype.includes) {
+ String.prototype.includes = function(value){
+   return this.indexOf(value)>-1
+ }
+}
+
+if (!String.prototype.includes) {
```
dkinzer: This should be `!Array.prototype.includes`

dkinzer:
@sensei100, one other thing  I noticed that the recommended Polyfill is more robust than this implementation, especially for the Array class.  Should we nut just copy those as our Polyfills instead of trying to implement our own versions?  Otherwise these functions will behave differently in e11 etc than they do where they are implemented natively @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

sensei100:
@dkinzer Sorry about the copy/paste mistake.  We are using the stimulus polyfills package and that handles most of what we need.  This was one method that was not included in that package and I read that it isn't included in the babel-polyfills package either.  I thought it was better to just handle this one error that isn't covered by our current package here rather than adding another whole package.

dkinzer:
OK, but just to be clear.  I don't mean adding a package I mean copy-pasting the recommended implementation of the actual function.

sensei100:
@dkinzer I'm not sure now where I got the example that I am using.  I liked that it was a little less code than the one you suggested.  We are using includes in only 1 place in the code, so I initially tried to replace it with indexOf to avoid dealing with it, but that was giving strange behavior form some of the records.  Changing it to this function seemed to work in Safari, so that is why I went with it.  I can use the other one if you think that is best.

dkinzer:
Note thought that the one with more code is a closer implementation of the actual protocol definition: https://tc39.github.io/ecma262/#sec-array.prototype.includes

And that protocol talks about how includes is supposed to be different than `IndexOf`:
```
NOTE 3
The includes method intentionally differs from the similar indexOf method in two ways. First, it uses the SameValueZero algorithm, instead of Strict Equality Comparison, allowing it to detect NaN array elements. Second, it does not skip missing array elements, instead treating them as undefined.
```
So using `indexOf` as the underlying implementation as we are doing might have some unwanted consequences (I don't know what they are since I'm not a javascript expert).

But like you said, if we are only using it in one place and in a very specific way that does not care about the difference between indexOf and includes, then maybe it's OK.

PR #639, opened 2018-08-30
Show only 1 aeon button

sensei100:
Fixes bug where multiple aeon buttons were displaying on record pages with multiple items


sensei100:
@bibliotechy I changed it to use any, but still need to call first on the items in order to get the single item information for the url. This bug was strange because it was not happing in the search results, only on the individual record pages.

PR #644, opened 2018-08-31
Change order that data-targets are called

sensei100:
Change order that targets are called to try to get rid of javascript console errors


dkinzer:
I'm still getting the error in chrome but I'm not getting it in firefox.

sensei100:
@dkinzer I made some more updates, so hopefully that helps.  It's hard to troubleshoot since I don't see any of these console errors locally.

dkinzer:
Sweet! that fixed it!

PR #646, opened 2018-08-31
Bl 637 advanced buttons

sensei100:
BL-637 Changes requested for advances search buttons
- Reverse order
- Update text to "Clear Form" and add new styling
screen shot 2018-08-31 at 3 52 12 pm


dkinzer:
Looks great!

PR #647, opened 2018-09-04
BL-629 subject_display subfields indexed together

sensei100:
BL-629 Subject display fields were getting the data mixed up
- Index named fields and description fields together so that the description fields don't go to the wrong line

### DIFF
--- a/lib/traject/indexer_config.rb
+++ b/lib/traject/indexer_config.rb
@@ -157,10 +157,7 @@ to_field "note_local_display", extract_marc("590a")
 
 # Subject fields
 to_field "subject_facet", extract_marc("600abcdefghklmnopqrstuxyz:610abcdefghklmnoprstuvxy:611acdefghjklnpqstuvxyz:630adefghklmnoprstvxyz:648axvyz:650abcdegvxyz:651aegvxyz:653a:654abcevyz:655abcvxyz:656akvxyz:657avxyz:690abcdegvxyz", separator: " — ", trim_punctuation: true)
-to_field "subject_display" do |record, accumulator|
-  subjects = process_subjects(record, "600abcdefghklmnopqrstuvxyz:610abcdefghklmnoprstuvxyz: 611acdefghjklnpqstuvxyz:630adefghklmnoprstvxyz:648axvyz:650abcdegvxyz:651aegvxyz:653a:654abcevyz:655abcvxyz:656akvxyz:657avxyz:690abcdegvxyz")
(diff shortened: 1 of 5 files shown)


lib/traject/indexer_config.rb
```diff
@@ -157,7 +157,10 @@
 
 # Subject fields
 to_field "subject_facet", extract_marc("600abcdefghklmnopqrstuxyz:610abcdefghklmnoprstuvxy:611acdefghjklnpqstuvxyz:630adefghklmnoprstvxyz:648axvyz:650abcdegvxyz:651aegvxyz:653a:654abcevyz:655abcvxyz:656akvxyz:657avxyz:690abcdegvxyz", separator: " — ", trim_punctuation: true)
-to_field "subject_display", extract_subject_display
+to_field "subject_display" do |record, accumulator|
```
dkinzer: @sensei100 please move this block to a function inside of custom.rb and add test for the expected behavior, thanks!

sensei100:
@dkinzer I made the changes you requested.

spec/lib/traject/traject_indexer_spec.rb
```diff
@@ -603,14 +603,6 @@
       end
     end
 
-    context "when a record doesn't have subject topics" do
-      let(:path) { "subject_topic_missing.xml" }
-
-      it "does not map anything to the field" do
-        expect(subject.map_record(records[0])).to eq({})
```
dkinzer: @sensei100 please add back a test for when there are no subjects so we don't inadvertently create a bug if we refactor this method and forget to account for records that do not have any subjects.  Thanks!

sensei100:
Per Leanne, there would never be a case where a subfield would only have v.x.y.z.  That test does break with this new method, but since it isn't necessary I removed it.

spec/lib/traject/traject_indexer_spec.rb
sensei100: @dkinzer That test doesn't really seem to be testing for no subjects.  The sample data includes a data field with subfields v and z.

dkinzer:
@sensei100 sorry, I'm not following. I'm not asking to add back that specific test you removed, just to add a case for if field is missing.  Such a case may arise (for example) under dev conditions when I create a dummy record and ingest it but I do not include any subjects.  Thanks!

dkinzer:
Sweet! I'll merge as soon as PR greens.  Thanks! 🙏 
🍬  🍬 🍬

PR #648, opened 2018-09-04
Add javascript testing.

dkinzer:
Adds the Jest JavaScript testing framework to our project with some initial tests and a documentation.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,5 +1,6 @@
 {
   "presets": [
+    "es2015",
     ["env", {
       "modules": false,
       "targets": {
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ before_script:
 script:
 - bundle exec rubocop
 - RELEVANCE=y bundle exec rake ci
+- yarn test
 notifications:
   slack:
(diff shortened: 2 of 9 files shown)


sensei100:
@dkinzer I'm just wondering what made you choose Jest over some of the other options out there since we aren't using react.

dkinzer:
@sensei100 I chose it because the framework is not actually react specific and all the blog post I read pretty much said that this one is the current champion.

dkinzer:
imho, they are all a pain to work with.

sensei100:
@dkinzer I was just curious.  I haven't written javascript tests before, but have used Jasmine and Mocha.  This site talked about AVA and Tape, which I haven't heard of before.  I just wondered what your process was for deciding on one. https://raygun.com/blog/javascript-unit-testing-frameworks/

dkinzer:
Yeah, my plan was to try a bunch and see which one I liked.  But then I spent all of Friday getting Jest to work with our setup.  👿 

If you have some good experience with those other frameworks we could give them a try too.  I have bad experiences with casperjs/phantomjs and selenium based tests (So I wouldn't recommend them).

bibliotechy:
This sees like a good idea, though I have nothing to add in terms of what test framework to use. @sensei100 , since you've been doing most of the JS lately, I'll leave this up to your discretion.

sensei100:
@dkinzer @bibliotechy My only hesitation in using this is that because it is relatively new, there isn't as much documentation and resources when we get stuck. That was a challenge with stimulus as times.  Jasmine has been around a long time and has more to rely on, but I have never actually written tests in it.

dkinzer:
@sensei100, if it helps, in terms of syntax jest and jasmine are exactly the same.  In fact jest used to use jasmine underneath. (I'm not 100% sure it does anymore but I know the syntax is still the same).

Also, Jest has been around since 2014.  So it's not entirely new.  And and added benefit is that if we ever want to incorporate some React components then Jest is better equipped to test those.

dkinzer:
I do agree that Stimulus has an issue with bad documentation.  In fact the hardest part of this PR was figuring out how to write a test for Stimulus.  Though I think I would have had some similar issues had I used another testing framework.

Stimulus uses qunit as their testing framework.  I didn't go with that because it seemed harder to set up.

PR #651, opened 2018-09-05
Add range-limit query.

dkinzer:
REF BL-350

Adds range limit query date facet.


dkinzer:
Thanks!  I thought I had accounted for that by not setting the facet unless both min and max are define but perhaps that bit isn't working as expected.  Looking into it.

dkinzer:
@bibliotechy I'v made the requested change.  Thanks! 🙏

app/models/blacklight/primo_central/search_builder_behavior.rb
```diff
@@ -105,7 +105,36 @@ def add_query_facets(primo_central_parameters)
...
+      primo_central_parameters[:range] = range
+
+      primo_central_parameters[:query][:q].facet(
+        field: "searchcreationdate",
+        value: "#{range}",
+      )
+    end
+
     private
+      class YearRange
```
bibliotechy: Does understanding how Primo expects dat ranges to be formated belong is the primo gem itself, not here?

dkinzer: Cool, yeah I could do that.  What do you think works betters as the interface?

```
Primo.facet(field: "fieldname", value: { range: min: 0, max: 0 })

# or

Primo.range_facet(field: "fieldname", min: 0, max: 0)
```

Or do you have something else in mind?

bibliotechy: Are there other fields that can be a range besides date?

dkinzer: I don't know 😸 ; actually it's not even documented that `searchcreationdate` is the field to do this with.  In fact that field isn't even documented on the api page.  They have a `fromDate` field documented but it doesn't take a range.

I learned about `searchcreationdate` from a forum post.

PR #658, opened 2018-09-07
Bl 628 pickup locations

sensei100:
BL-628 Items on reserve or reference, but available at another location should be able to be picked up at library that has reserve copy
- 991033204029703811 should show Paley Library in the pickup locations dropdown, but not Remote Storage


app/lib/cob_alma/requests.rb
```diff
@@ -53,15 +53,35 @@ def self.avail_locations(items_list)
...
+      current_libraries = items_list.collect { |k, v|
+        v.map { |item|
+          [item.item_data.dig("location", "value")]
+        }.flatten }
+      locations = items_list.keys.zip(current_libraries).to_h
+
+      if locations.length > 1
+        pickup_locations = []
+        reserve_or_reference = locations.each.map { |k, v|k if v.all? { |i| i == "reserve" || i == "reference" } }.compact
+        additional_unreserved_copy = locations.each.map { |k, v|k if v.all? { |i| i != "reserve" || i != "reference" } }.compact
```
dkinzer: Hi @sensei100, there are a several issues going on here I'll list them below:

* The predicate for `additional_unreserved_copy` is alway true as written:
```
 i != "reserve" || i != "reference"
```
Anything `i` can be makes the above true:
```
i == foo, then true because i != reverse.
i == reverse, then true because i != reference
i == reference, then true because i != reverse
```
Thus `reserve_or_reference && additional_unreserved_copy` can only be false when `reserve_or_reference`.

I've verified locally that not using `additional_unreserved_copy` keeps your tests passing.

* Next, `locations.each.map ...` is redundant it should just be `locations.map`

* The result of a map is always an `Array`, and arrays are always truthy thus `if reserve_or_reference` will always be true:  I have verified locally that your test continue to pass if I remove `if reserve_or_reference && additional_unreserved_copy`

* Finally, It might be cleaner to use `.reduce` instead of `.map` in this case (this one I'm not too sure about, so feel free to disregard).

sensei100:
@dkinzer I've refactored.  Let me know if you think the changes will work.

dkinzer:
Thanks!

PR #660, opened 2018-09-07
Update limit on article OR queries.

dkinzer:
The new api has a limit of 13 vs 9 OR queries. This change reflects that
change.


dkinzer:
This should probably be refactored to go into Primo too.. So primo handles making multiple search queries and knows about the limits.

PR #665, opened 2018-09-11
Bl 571 ris files

sensei100:
BL-571 Use blacklight-ris gem to enable the user to download ris files
- @dkinzer Can we talk about this approach?


dkinzer:
@sensei100 sure thing

PR #672, opened 2018-09-13
Revert "Remove css that is breaking libqa. (#671)"

dkinzer:
This reverts commit e0cc63602f396ae03b5feec616c3b328bdb88b91.

(yup, that didn't fix anyting.... maybe it was the manual deploy)


sensei100:
@dkinzer Chris said the -ms css methods aren't reliable and shouldn't be used in production.  Maybe we should leave it out.

PR #682, opened 2018-09-17
Bl 571 isbn export

sensei100:
BL-571 SN tag was only displaying issn OR isbn
- Use block to export both fields

### DIFF
--- a/app/models/solr_document.rb
+++ b/app/models/solr_document.rb
@@ -69,7 +69,7 @@ class SolrDocument
     ET: "edition_display",
     LA: "language_display",
     KW: "subject_display",
-    SN: Proc.new {self["isbn_display"] || self["issn_display"]},
+    SN: Proc.new { self["isbn_display"] || self["issn_display"] },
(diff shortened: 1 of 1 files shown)


app/models/solr_document.rb
```diff
@@ -69,8 +69,7 @@ def initialize(doc, req = nil)
     ET: "edition_display",
     LA: "language_display",
     KW: "subject_display",
-    SN: "isbn_display",
-    SN: "issn_display",
+    SN: Proc.new { self["isbn_display"] || self["issn_display"] },
```
dkinzer: It's too bad the blacklight-ris gem developers chose to represent the mappings with refwork tag as the key instead of the solr field name as the key.

When I implemented that mapping for refworks show tool I switched the order of the mapping so that multiple fields could point to the same tag:

https://github.com/tulibraries/tul_cob/blob/master/app/models/blacklight/primo_central/document_export.rb#L44

Maybe we should contribute a PR to that gem.

PR #695, opened 2018-09-20
Remove search form from bookmarks page.

dkinzer:
REF BL-656


sensei100:
@dkinzer When I click on the file, it says there are no changes.

dkinzer:
that's because it's an empty template file 😀   that's all that is required in this case.

PR #711, opened 2018-09-25
Tweak the solr document irs tags.

dkinzer:
Use :Y1 instead of :YR for the date tag.


sensei100:
@dkinzer It is the PY tag.

dkinzer:
cool, thanks!  was just about to add "pub_date_sort" :D

PR #720, opened 2018-09-27
Updates article branding.

dkinzer:
REF BL-657

Also:
* refactors bento_titleize helper method.
* Adds tests for bento_titleize helper method.


sensei100:
The decision was made to leave this as is until user testing can be done.

PR #736, opened 2018-10-04
BL-667 pickup locations by item

sensei100:
Our former logic was only working at the bib level, so patrons could not request a specific volume that was not available at their library.  
- Refines the current valid pickup locations method to get rid of a bug that was 
including locations that shouldn't be there  
- Adds logic to filter pickup locations by description at the item level.


dkinzer:
Very good!

PR #741, opened 2018-10-10
Add some more synonyms to solr config.

dkinzer:
REF BL-695

Add the following synonyms to Solr search:
Cantatas = kantate
Concerto = Konzerte
Mass = Messe
Sonatas = sonaten
Symphonies = sinfonien
Etc. = etcetera
Piano = Klavier
symphony = Sinfonie
arr. = arranged
maj. = major

### DIFF
--- a/solr/conf/synonyms.txt
+++ b/solr/conf/synonyms.txt
@@ -52,7 +52,19 @@ c++ => cplusplus
 
 min => minor
 maj => major
+maj. => major
 
 # Added for BL-447
 employment,labor => work
 precarious => contingent
+
+# Added for BL-695
+kantate => cantatas
+konzerte => concerto
+messe => mass
+sonaten => sonatas
+sinfonien => symphonies
+etc. => etcetera
+klavier => piano
+sinfonie => symphony
+arr. => arranged


sensei100:
@dkinzer The previous synonyms in this file all use a different syntax.  Does either way work?

dkinzer:
@sensei100  there are three alternative syntax with slightly different meanings: https://doc.lucidworks.com/fusion/3.0/Collections/Synonyms-Files.html

 I put in a question to at @ebtoner on ticket about it.  I chose the one that gives them all equal weight; but we can choose another way if it makes more sense.

bibliotechy:
That looks like docs for Fusion, an entirely different product.

The two syntaxes do have different implications as per [solr docs]( https://lucene.apache.org/solr/guide/6_6/filter-descriptions.html#FilterDescriptions-SynonymGraphFilter)

"There are two ways to specify synonym mappings:

A comma-separated list of words: If the token matches any of the words, then all the words in the list are substituted, which will include the original token.

Two comma-separated lists of words with the symbol "⇒" between them: If the token matches any word on the left, then the list on the right is substituted. The original token will not be included unless it is also in the list on the right."

In my understanding, the one you choose also ends up having an effect on relevance as inserting multiple extra terms into a query or into the index has implications for term distance calculations.

The approach I took earlier is to always replace any terms with it's preferred term at both query and index time, so that we get the synonym substitution, but don't skew term distance relevancy scoring.

dkinzer:
oops sorry about giving the url to the wrong product (amazingly they seem to be using very similar substitution technique).

Anyway, it's still not clear to me which one makes the most sense.  Using the free list still gives all values the same weight and since I don't know what is the preferred value in these cases .... like between Cantatas vs kantate, does it not make sense to treat them all equally?

bibliotechy:
So, in the substitution model already in use, `cantatas => kantate`, when solr sees the term `cantatas` in a record being added to the index, it will replace it with `kantate`. Similarly, when a user searches for `canatas`, solr will substitute the term `kantate`. They get equal weight because solr is treating them as the same term and query and index time. 

Neither is preferred in the sense of one being weighted higher than the other, just that one is chosen as the term we will ue in the index and query substitution, which is invisible to users, and to blacklight

dkinzer:
Hmm, ok.  It almost sounds like it doesn't matter what we use then, the result is the same.

PR #745, opened 2018-10-10
Makes the bento search the root of the site.

dkinzer:
REF BL-680

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -47,7 +47,7 @@ module ApplicationHelper
 
   def subject_links(args)
     args[:document][args[:field]].map do |subject|
-      link_to(subject, "#{root_path}?f[subject_facet][]=#{CGI.escape subject}")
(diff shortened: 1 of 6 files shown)


sensei100:
@dkinzer I think this will affect links that include the root path.  I know that subject links refer to that. I think that others might as well.  Do we need to change all of those?

dkinzer:
@sensei100 roo_path will be "/" this only afffects the controller and action that queries to root get sent to.  I expected some tests to fail that depend on going to "/" and doing a catalog search.  I'm happy to see only feature tests are failing this way.  Will fix.

sensei100:
@dkinzer This does break some current links.  When I click on a subject link in the record page, it now takes me to an empty bento search.

dkinzer:
@sensei100 good catch, let's fix those links

dkinzer:
@sensei100 ok, I updated the helper method so that it doesn't assume that root_path is where catalog_search should be happening.

PR #747, opened 2018-10-11
Normalize the bento search (everything) with nav bar.

dkinzer:
Currently the links and searches switching between everything and other
searches does not propogate.  This change propagates that search so that
flipping back and forth between a facted books search and everything
returns the same results (per example.)


dkinzer:
closing as looks like we wont need this

PR #760, opened 2018-10-16
[BL-689] Add breadcrumbs to advanced pages.

dkinzer:
REF BL-689

Use breadcrumbs_on_rails gem to add breadcrumbs to the advanced pages
plus books and journals.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -94,6 +94,7 @@ gem "browser"
 gem "blacklight-ris", git: "https://github.com/upenn-libraries/blacklight-ris.git"
 gem "bootstrap-select-rails"
 gem "httparty"
+gem "breadcrumbs_on_rails"
 
 group :production do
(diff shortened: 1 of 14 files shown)


sensei100:
@dkinzer Will this save the search terms and send them back to their original search state?

dkinzer:
not yet, is that a requirement?

sensei100:
@dkinzer Yes, both Jira tickets asked for the breadcrumb to retain their current search.  Using the link_back_to_catalog blacklight link worked for all controllers to save state.

dkinzer:
Cool, breadcrumbs don't usually save state like that but we can pass any method we like so that can happen.  I'm going to put this aside for now and get back to it later so I can fix the issue with missing bib items that needs to go out on the next prod deploy.

PR #763, opened 2018-10-17
BL-545 EZ-borrow integration

sensei100:
Uses the alma request options api to retreive the link to EZ-borrow.  
- The link should only appear on items that are books currently unavailable at Temple Libraries.


dkinzer:
@sensei100 This looks great!  And now that I'm looking at it closer I don't think it needs any further refinements.

PR #782, opened 2018-10-29
BL-718 Book requests with empty descriptions

sensei100:
Previous code was excluding a single empty element, but some items have multiple empty descriptions.  Changes code to remove all empty elements and adds tests for the method.

### DIFF
--- a/app/lib/cob_alma/requests.rb
+++ b/app/lib/cob_alma/requests.rb
@@ -116,7 +116,7 @@ module CobAlma
           desc
         end
       end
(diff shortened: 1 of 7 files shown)


app/lib/cob_alma/requests.rb
```diff
@@ -116,7 +116,7 @@ def self.descriptions(items_list)
           desc
         end
       end
-      descriptions unless descriptions == [""]
+      descriptions.reject(&:empty?)
```
dkinzer: 👍

PR #785, opened 2018-10-30
BL-699 Request modal should default to being open 

sensei100:
- when only hold requests are allowed


dkinzer:
@sensei100, do you mind adding one test case for when it passes.  Thanks!

dkinzer:
Oops I see the test case is there.  All good 👍

PR #794, opened 2018-11-02
BL-431 Updates to imprint_display

sensei100:
- Changes mappings for imprint_display
- Adds imprint_prod, imprint_dist, and imprint_man fields
- Changes date_copyright mappings

### DIFF
--- a/spec/features/record_page_fields_spec.rb
+++ b/spec/features/record_page_fields_spec.rb
@@ -184,14 +184,6 @@ RSpec.feature "RecordPageFields" do
       end
     end
 
-    let (:item_264_0) { fixtures.fetch("imprint_264_0") }
-    scenario "User visits a document with imprint indicator2 value 0" do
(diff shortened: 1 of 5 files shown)


spec/features/record_page_fields_spec.rb
```diff
@@ -184,70 +184,16 @@
       end
     end
 
-    let (:item_264_0) { fixtures.fetch("imprint_264_0") }
```
dkinzer: I only see changes above for field 260, should these 264 field tests be kept?

sensei100: @dkinzer No, most of the 264s were moved into other fields. I was talking with Chad about whether to keep writing feature specs for record page displays and he felt like it wasn't necessary unless the field had complicated logic.  That's why I took this out and didn't refactor into other specs.

dkinzer: ok, thanks.

PR #801, opened 2018-11-06
BL-359 Refactor availability display

sensei100:
Currently, if an item is suppressed it is still displaying the library name with an empty table.  This filters it to show the default text when all items are suppressed.

### DIFF
--- a/app/views/almaws/item.html.erb
+++ b/app/views/almaws/item.html.erb
@@ -1,12 +1,12 @@
 <button type="button" class="ajax-modal-close close" data-dismiss="modal" aria-hidden="true">×</button>
 
 <div id="request-url-data-<%= @mms_id %>" class="hidden" data-requests-url="<%= request_options_path(@mms_id, @pickup_locations, @request_level) %>"></div>
-<% unless @items.empty? || @unsuppressed_items.values.include?([]) %>
+<% unless @items.values.all?(&:empty?) %>
   <div class="availability-modal">
     <div class="modal-only">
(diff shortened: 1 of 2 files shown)


app/views/almaws/item.html.erb
```diff
@@ -1,8 +1,7 @@
 <button type="button" class="ajax-modal-close close" data-dismiss="modal" aria-hidden="true">×</button>
 
 <div id="request-url-data-<%= @mms_id %>" class="hidden" data-requests-url="<%= request_options_path(@mms_id, @pickup_locations, @request_level) %>"></div>
-
-<% unless @items.empty? %>
+<% unless @items.empty? || @unsuppressed_items.values.include?([]) %>
```
dkinzer: Sorry @sensei100 , but I'm not quite following what is going on here.  `@items` gets mutated in place by `helpers.unsupressed_holdings` so @items.empty doesn't  quite make sense, since it would only be empty if there are no locations.

`@unsupressed_items.values.include?([])` is the same as `@items.values.include?([])` . So why use a completely new variable for this?

The other thing that is odd is that this commit says
> This filters it to show the default text when all items are suppressed.
But `@unsupressed_items.values.include?([])` returns true when any location in @items has an empty set, not just the location you want to have a default value for.  So this will work when there is only one location to consider but it will fail when there are multiple locations one of which has all items suppressed.

dkinzer:
```suggestion
<% unless @items.values.all?(&:empty?) %>
```

PR #805, opened 2018-11-08
BL-690 Copy changes and styling updates

sensei100:
for search results displays and home pages


app/views/catalog/_search_results.html.erb
```diff
@@ -21,16 +15,16 @@
...
 <% end %>
 
-<%= render 'search_header' %>
+<%= render "search_header" %>
 <span id="alma_availability_url" data-url="<%= alma_availability_url(format: :json) %>" ></span>
-<h2 class="sr-only"><%= t('blacklight.search.search_results') %></h2>
+<h2 class="sr-only"><%= t("blacklight.search.search_results") %></h2>
 
-<%- if @response.empty? %>
+<%- if @response.empty? || empty_article_response?(@response) %>
```
dkinzer: Let's change empty_article_response? to empty_response? and take care of both of these cases.

```suggestion
<%- if empty_response?(@response) %>
```

app/views/catalog/_search_results.html.erb
```diff
@@ -1,15 +1,9 @@
-<h2 class="sr-only top-content-title"><%= t('blacklight.search.search_results_header') %></h2>
-
-<% if params[:controller] == "catalog" %>
```
dkinzer: Cool, I like seeing logic get removed from the template 🎉 😎

app/helpers/application_helper.rb
```diff
@@ -298,7 +298,22 @@ def former_search_link
   end
 
   def help_link
-    link_to t("ask_librarian"), Rails.configuration.ask_link
+    link_to t("ask_librarian"), Rails.configuration.ask_link, target: "_blank"
+  end
+
+  def explanation_translations(controller_name)
```
dkinzer: I've been meaning to move these controller metadata helpers directly in the controller (i.e. `BooksController.explenation_translation` etc.).  Let's not worry about it now, but just mentioning it as something we'll want to take on in the future.

PR #834, opened 2018-11-26
BL-620 Clean up css files

sensei100:
- Create/reorganize related styles into partials
- Removes unused css styles
- Ensure that color vars are being used


dkinzer:
Looks good.  I'm my local info tex for everything view is squeezed together now, but the issue is also present in liqa, so  it's not being caused by this.

PR #841, opened 2018-11-29
[BL-347] Refactor electronic collection/service info.

dkinzer:
REF BL-347

* Refactors link index to use JSON.
* Refactors solr config to parse new JSON fields.
* Refactors link helper methods to use new link objects.
* Refactors link to render collection/service notes.
* Updates blacklight to version 6.18.0


bibliotechy:
👏 👏 👏 
So happy to see this work. I always wanted to implement those, but never quite got it working.

dkinzer:
@bibliotechy thanks for letting me know about that solr feature!  It's so neat.

PR #851, opened 2018-12-05
BL-747 Remove item limits from request descriptions

sensei100:
Currently we are limited to only 100 item descriptions, so if a patron wants to request something from a larger set of records, they may not be able to choose the item they want.  This now returns all items in the request dropdown menus.

### DIFF
--- a/app/lib/cob_alma/requests.rb
+++ b/app/lib/cob_alma/requests.rb
@@ -109,8 +109,7 @@ module CobAlma
     end
 
     def self.descriptions(items_list)
-      descriptions = items_list.map { |item| item["item_data"]["description"] }
-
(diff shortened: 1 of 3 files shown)


app/lib/cob_alma/requests.rb
```diff
@@ -109,8 +109,7 @@ def self.equipment(items_list)
     end
 
     def self.descriptions(items_list)
-      descriptions = items_list.map { |item| item["item_data"]["description"] }
-
+      descriptions = items_list.all.map(&:description)
```
dkinzer: I ❤️ this.  We know these  items have a description without needing to know any further details.

PR #852, opened 2018-12-05
[DevOps] Allow multiple ingest of the same file.

dkinzer:
Now that we have configured solr to check the update date before
ingesting a file, it has been necessary locally to dump our solr
document before doing a re-ingest.  This step can become cumbersome if
we need to ingest the same file many times over during a development
session.

This commit adds a way to override this feature by setting the update
date to the current Time, thus guarantying that the update date is
always newer and effectively disable the Solr feature.


relaxing:
This looks good 👍

PR #861, opened 2018-12-10
Individual ids for each type of form 

sensei100:
were preventing the forms from working properly.  Removes ids assigned to select tags.


sensei100:
I changed my mind on the best way to handle this issue.

PR #882, opened 2018-12-18
BL-772 Update filters and facets for responsive views

sensei100:


app/assets/javascripts/application.js
```diff
@@ -146,4 +146,8 @@ $(document).on('turbolinks:load', function() {
  	if ($("div.navbar-form").length == 0) {
  		$("#search-navbar").css("padding-left", "15%");
  	}
+
+	$('#facet-filter-icon').click( function(){
+    $(this).find('span#facet-icons').toggleClass('open-facet-icon').toggleClass('remove-facet-icon');
```
dkinzer: @sensei100 quick question on this, why is this needed? How can I test it?  Thanks!

sensei100: @dkinzer They want a different icon to be displayed when the facets are open/closed in the responsive views.  You can see it by using the responsive dev tools in chrome and looking at the yellow icon for the facets.  It should be the filter icon when facets are closed, and show the close icon when they are open.

dkinzer: Thanks! 👍

PR #892, opened 2019-01-03
Minor performance updates

sensei100:
- Resize instagram image
- Defer js files
- Moves all header tags into one layout file


dkinzer:
I think we can just get rid of `app/views/layouts/application.html.erb` everything just seems to use the blacklight layout.

PR #904, opened 2019-01-15
BL-784 Item level request options

sensei100:
- We are not always receiving the correct request options for item level requests. This sometimes results in a user not being offered any request options, even when they should be available.
- This will iterate over each holding_id/item_pid until it finds a combination that does have request options. 
- This is a short term fix for the problem until we can meet and decide on how we want requests to work/look going forward.


app/controllers/almaws_controller.rb
```diff
@@ -51,7 +49,16 @@ def request_options
     @request_level = params[:request_level]
     start = Time.now
     if @request_level == "item"
-      @request_options = Alma::ItemRequestOptions.get(@mms_id, @holding_id, @item_pid, user_id: @user_id)
+      @item_level_holdings = CobAlma::Requests.item_holding_ids(@items)
```
dkinzer: @sensei100 this looks good, but breaking out of the each block should be done using `break` for example:

```ruby
[1, 2, 3].each { |n| @n = n; break if n > 1; }
```
Could you please refactor this section by removing `keep_going` references and use a `break` statement as per example above?

Thanks!

PR #930, opened 2019-01-25
BL-789 xml views for books and journals

sensei100:
- xml views and refworks are not currently available for books and journals. 
- Adds the refworks_marc_txt to each model 
- Adds helper methods for the refworks links


dkinzer:
Great job! 💯  I guess I was wrong about those paths being generated by the router config 😊

PR #933, opened 2019-01-28
Bl 694 refworks bookmarks

sensei100:
- Adds back a helper method accidentally removed last week


dkinzer:
🤔 wonder if this will also fix issue with blacklight-7 upgrade and doc refworks?

sensei100:
@dkinzer I'm not sure which issue you mean.  I thought we fixed bookmarks last week.

PR #948, opened 2019-02-04
[BL-348] Implement Content DM API integration.

dkinzer:
REF BL-348

 ## Requirements

 * In the "Resource Type" column of the bento box, add Resource type
 value for "Digital Collections"

 * This value should include the number of results that there are for
 that query in ContentDM, accessed via ContentDM API

 * The facet link itself should open the search results for ContentDM,
 with the basic query from the Library Search.

### DIFF
--- a/spec/models/search/solr/response_spec.rb
+++ b/spec/models/search/solr/response_spec.rb
@@ -28,7 +28,7 @@ RSpec.describe Search::Solr::Response, type: :model do
       end
     end
 
-    context "facet exits and field exists" do
+    context "facet exists and field exists" do
       let(:facet) { { name: "cat", value: "memory", hits: 4 } }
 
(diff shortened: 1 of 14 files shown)


spec/models/search/solr/response_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+
+    context "facet exists but field does not exist" do
+      let(:facet) { { name: "cat", value: "bar", hits: 1 } }
+
+      it "appends the new field name and value" do
+        expect(response.facet_fields["cat"]).to eq(["memory", 3, "card", 2, "bar", 1])
+      end
+    end
+
+    context "facet exits and field exists" do
```
sensei100:
```suggestion
    context "facet exists and field exists" do
```

PR #971, opened 2019-02-14
Solr timezone issue

relaxing:
This solrconfig change addresses the issue with timezones maybe or maybe not appearing in the MARC XML from Alma.


solr/conf/solrconfig.xml
```diff
@@ -491,17 +491,18 @@
 
   <updateRequestProcessorChain name="DocCentricVersioningOnDate">
     <processor class="solr.ParseDateFieldUpdateProcessorFactory">
+      <str name="defaultTimeZone">Etc/UTC</str>
```
dkinzer: @relaxing can you please explain why we are using this as our default time zone. Does this match the default for Alma marc dump which I thought was UTC - 5 hours?

relaxing: The old Alma dumps were UTC. Then they turned into US/Eastern (UTC+5).

dkinzer: OK, I guess what I'm looking for is how will this fix our issue, what the issue is, so I can document it in the commit.  Also I'm confused because if Alma now defaults to "UTC -5" wont defaulting solr to Etc/UTC when zone is not present cause a 5 hour diff in actual time.

relaxing: My assumption was that if they suddenly stop appending a time zone, it's because they went back to the old behavior (UTC).

dkinzer: OK, thanks!

PR #984, opened 2019-02-15
Initial solr replication request handler

relaxing:
Before merging we need to have our deploy scripts put the correct solrconf fragment in place.


bibliotechy:
So...what would you see the deployment story being on this? Copy the config over to both solr's, if it is a follower, delete the leader XML?

relaxing:
> So...what would you see the deployment story being on this? Copy the config over to both solr's, if it is a follower, delete the leader XML?

Yeah, that seems like the easiest way to do it from Ansible.

relaxing:
Also put the right IP address and core name into the follower config.

dkinzer:
BL-450 is marked done on Jira so closing this.

PR #1016, opened 2019-02-26
Update factory_bot_rails gem, revert sqlite3

sensei100:


dkinzer:
This one can be closed too as we decided not to pursue these updates currently and the rebase of blacklight-7 is causing all these commits to re-apply.

PR #1017, opened 2019-02-26
Revert gems

sensei100:


dkinzer:
Looks like something is off with this... let's redo  it.

PR #1028, opened 2019-02-28
Enable suggestions for solr searches.

dkinzer:
REF BL-316

* Makes suggestions on entire phrase (not just per word)
* Max suggestions is 3 (default is 1)
* Disables suggestions for advanced search (does not work)


sensei100:
@dkinzer I was only able to get this working locally after doing a reindex, so I added the required reindex label.  Looks good!

dkinzer:
@sensei100 btw this only requires that we run optimize ... a full reindex is not required (but almost same diff I guess ... less work though 💃 ).

PR #1029, opened 2019-03-01
Fix send to email from books and journals.

dkinzer:
REF BL-822

Send to email is broken from books and journal page because those
document classes are not being extended properly.

Since these classes should act as aliases to SolrDocument anyway,
updating them to equal SolrDocument fixes the issue.

### DIFF
--- a/app/models/solr_book_document.rb
+++ b/app/models/solr_book_document.rb
@@ -1,13 +1,3 @@
 # frozen_string_literal: true
 
-class SolrBookDocument < SolrDocument
-  use_extension(Blacklight::Document::Email)
-  use_extension(Blacklight::Document::Sms)
-
-  # Add Blacklight-Marc extension:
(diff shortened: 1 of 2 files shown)


sensei100:
@dkinzer Too bad we didn't think of this earlier when working on the bookmarks stuff!  Should we remove the helper methods that reference SolrBookDcoument and SolrJournalDocument now?

dkinzer:
@sensei100  I don't think there are any helper methods that reference them.  But, yeah... it's definitely related.  I mean really I'm thinking we could probably just do away with the aliases and just use SolrDocument model for all.

But if go this route we'll have to do a migration too to fix data in production ... What do you think?

dkinzer:
Closing  to pursue just using SolrDocument in all the places.. should fix this and also issues we have been seeing with bookmarks.

PR #1056, opened 2019-03-15
BL-835 Remove dropdown from bento view

sensei100:
- The search fields select menu was accidentially introduced to the bento view.  It doesn't work properly, so it should be hidden for bento searches

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -4,6 +4,7 @@ class SearchController < CatalogController
   include CatalogConfigReinit
 
   blacklight_config.configure do |config|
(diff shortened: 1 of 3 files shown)


dkinzer:
That dropped down is controlled via configuration and I think I inadvertently enabled it when I was doing cdm work https://github.com/tulibraries/tul_cob/blame/master/app/controllers/search_controller.rb#L7.  We should disable it by removing the search_fields configured for this controller.

sensei100:
@dkinzer I took out the search_field config on line 7, but the dropdown is still visible in the bento search.  Can you think of another place where it needs to be taken out?

PR #1076, opened 2019-03-25
BL-844 Updates to availability_status method

sensei100:
- items with a physical_material_type of "Bound Issue" should display as "Libary Use Only"
- Items with a circulation_policy of "Music Restricted" should also be Library Use Only

### DIFF
--- a/app/helpers/alma_data_helper.rb
+++ b/app/helpers/alma_data_helper.rb
@@ -7,7 +7,9 @@ module AlmaDataHelper
 
   def availability_status(item)
     if item.in_place? && item.item_data["requested"] == false
-      if item.non_circulating? || item.location == "reserve" || item.physical_material_type.fetch("desc", "") == "Bound Issue" || item.circulation_policy == "Music Restricted"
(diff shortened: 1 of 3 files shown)


app/helpers/alma_data_helper.rb
```diff
@@ -7,7 +7,7 @@ module AlmaDataHelper
 
   def availability_status(item)
     if item.in_place? && item.item_data["requested"] == false
-      if item.non_circulating? || item.location == "reserve"
+      if item.non_circulating? || item.location == "reserve" || item.physical_material_type.fetch("desc", "") == "Bound Issue" || item.circulation_policy == "Music Restricted"
```
dkinzer:
```suggestion
      if item.non_circulating? || item.location == "reserve" ||
          item.physical_material_type.fetch("desc", "") == "Bound Issue" ||
          item.circulation_policy == "Music Restricted"
```

dkinzer: Let's break up this long or chain to make it easier to see on screen.

PR #1105, opened 2019-04-05
BL-846 Bento links

sensei100:
-- Ensure nav tools appear on record pages when coming from a bento search


app/views/catalog/show.html.erb
```diff
@@ -1,3 +1,5 @@
+<% @search_context ||= current_search_session %>
```
dkinzer: We're missing the comment to remind us why this is here.

sensei100: Weird.  I'll add it back in.

PR #1123, opened 2019-04-11
Show availability for bookmarks

sensei100:


app/views/catalog/_fields.html.erb
```diff
@@ -18,6 +18,6 @@
 <% end -%>
 </dl>
 
-<% if controller_name === "catalog" || controller_name === "books" || controller_name === "journals" %>
+<% if controller_name == "catalog" || controller_name == "books" || controller_name == "journals" || controller_name == "bookmarks" %>
```
dkinzer: At some point I'd like to add availability in a similar fashion to a tool field via configuration.  But for now maybe we can make this line a little more succinct... perhaps?

```suggestion
<% [ "catalog", "books", "journals", "bookmarks" ].include?(controller_name) %>
```

PR #1125, opened 2019-04-12
Add script for generating list of children mmsids.

dkinzer:
This script is used in the process of creating the boundwith_merged.xml file.


dkinzer:
@relaxing should I close this?

cmharlow:
@relaxing okay to close this? We won't delete, so we have the history of what scripts were helpful for boundwith processing in the current process; but I imagine this is being rewritten / refactored as part of your cob datapipeline DAG boundwith processing. Sound okay to close PR?

PR #1161, opened 2019-04-29
[BL_809] Fix no diff in databases subject vs title search.

dkinzer:
* Overrides search settings for databases az so that it does not use the 
default search settings which are currently optimized for catalog search.

* Adds az-database core to test and local via solr_wrapper.


dkinzer:
@relaxing @cmharlow ... pinging you for review of this one too.

PR #1172, opened 2019-05-03
[BL-860] Add ASRS specific request panel.

dkinzer:
* Adds additional request form for ASRS items.


sensei100:
@dkinzer Some of the specs were failing due to the fact that we are allowing MAIN to be a pickup location during the move.  I skipped them for now with comments explaining why they are skipped.

dkinzer:
@sensei100 thanks for fixing that

PR #1188, opened 2019-05-09
Fix syntax errors

sensei100:
- remove closing div tag from almaws/item
- add closing div to show_availability
- use html_safe characters for greater than symbols
-remove extra row class breaking IE


dkinzer:
Looks great!  I test locally on IE11 via VM

PR #1209, opened 2019-05-17
BL-892 Update location mappings

sensei100:
- Updates location mappings for all libraries 
- Requires reindex

### DIFF
--- a/config/locations.yml
+++ b/config/locations.yml
@@ -1,20 +1,24 @@
 default: &default
   AMBLER:
-    aarc: Ambler Archives Collection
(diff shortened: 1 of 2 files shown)


dkinzer:
@sensei100 should we also update:
./config/translation_maps/libraries_map.properties
./config/libraries.yml

Or if those aren't being used anymore we should get rid of them... (I think libraries_map.properties) is what traject uses.

sensei100:
@dkinzer I'm not sure what you are referring to. If it is the Media stuff, we are leaving it in for now.

PR #1218, opened 2019-05-22
BL-877 Availability error message

sensei100:
- Currently some availability statuses are empty, which may cause confusion for the patron.
- Adds an alert message to the top of the availability panel when an availability status is left empty.

### DIFF
--- a/app/helpers/availability_helper.rb
+++ b/app/helpers/availability_helper.rb
@@ -69,6 +69,11 @@ module AvailabilityHelper
       .group_by { |item| library(item) }
   end
 
+  def availability_alert(document)
+    empty_availability = document["items_json_display"].map { |item|
+      item["availability"].blank?
+    }.any?
+  end
 
   def description(item)
     item["description"] ? "Description: #{item['description']}" : ""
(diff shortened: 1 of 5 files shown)


app/helpers/availability_helper.rb
```diff
@@ -69,6 +69,11 @@ def document_and_api_merged_results(document, items_list)
       .group_by { |item| library(item) }
   end
 
+  def availability_alert(document)
+    empty_availability = document["items_json_display"].map { |item|
```
dkinzer: I don't think we need `empty_availability =` here since it's set and not used.

app/views/almaws/item.html.erb
```diff
@@ -3,6 +3,12 @@
 <%# This template has access to @items returned by the alma api, so it has slightly different information that the other template %>
 <%# Any changes related to the alma api should only be made here %>
 
+<% if availability_alert(@document) == true %>
```
dkinzer: This should just be `if availability_alert(@document)` since `availability_alert(@document)` will return `true` or `false`

sensei100: @dkinzer I made the changes you requested.

PR #1235, opened 2019-05-31
Fix for records with deleted fields without a corresponding date update.

relaxing:


dkinzer:
@relaxing I just submitted a PR against this one (#1236) to fix lintting issue and refactor.

PR #1245, opened 2019-06-06
BL-810 Database resource types

sensei100:
- Makes resource types searchable in all fields searches
- Create separate field for "database" type so that it doesn't display in the facets (currently not being used, but may be in the future)
- Align database type links of record pages

### DIFF
--- a/app/controllers/databases_controller.rb
+++ b/app/controllers/databases_controller.rb
@@ -26,6 +26,7 @@ class DatabasesController < CatalogController
         title_t^10000
         subject_facet^1000
         format^200
+        format_t^200
         note_t^100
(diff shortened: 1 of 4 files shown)


app/controllers/databases_controller.rb
```diff
@@ -103,7 +106,7 @@ class DatabasesController < CatalogController
       field.solr_parameters = { "spellcheck.dictionary": "subject" }
       field.qt = "search"
       field.solr_local_parameters = {
-        qf: "$subjec_qf",
+        qf: "$subject_qf",
```
dkinzer: Good catch!

PR #1333, opened 2019-08-01
Skip solr tasks to avoid privilege escalation error

bibliotechy:


cmharlow:
I don't think the issue is solr-related, though?

PR #1335, opened 2019-08-01
BL-935 Request form updates

sensei100:
- Updates background color for headers
- Text changes for Request headers
- Adds wait times for ASRS pickup locations
- Use Charles as default option for ASRS pickup locations
- Remove notes field from ASRS form


sensei100:
@cmharlow Is that the typo in the PR that you opened a few minutes ago? Or is there one in this PR as well?

cmharlow:
the repair is in the small PR I made against your branch.

PR #1399, opened 2019-08-26
Revert "BL-964 MAIN stacks should be requestable at Charles (#1381)"

dkinzer:
This reverts commit 6a763b223c7fdaaf57af1a52edece237e81eac32.


sensei100:
@dkinzer This reverts only one of the changes made. If you look at the Jira ticket, there were several commits related to this ticket that also need to be reverted.

PR #1411, opened 2019-08-30
BL-937 Use preferred names

sensei100:
- Retrieves preferred user names from Alma
- Changes text for directions to update name
- Moves longer text blocks on account_info page into locales file

### DIFF
--- a/app/views/users/_user_info.html.erb
+++ b/app/views/users/_user_info.html.erb
@@ -1,16 +1,16 @@
 <div class="user-info row row-eq-height">
 	<div class="col col-md-5">
 	  <ul class="user-info-list">
-	    <li><span class="user-info-label">Name:</span> <%= current_user.name %></li>
+	    <li><span class="user-info-label">Name:</span> <%= @user.alma.preferred_name %></li>
 	    <li><span class="user-info-label">Status:</span> Active </li>
 			<li><span class="user-info-label">Items requested:</span> <%= @user.alma.total_requests %></li>
 	    <li><span class="user-info-label">Items checked out:</span> <%= @user.alma.total_loans %></li>
 	    <li><span class="user-info-label">Current balance:</span>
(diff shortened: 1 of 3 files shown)


app/views/users/_user_info.html.erb
```diff
@@ -1,16 +1,16 @@
 <div class="user-info row row-eq-height">
 	<div class="col col-md-5">
 	  <ul class="user-info-list">
-	    <li><span class="user-info-label">Name:</span> <%= current_user.name %></li>
```
dkinzer: I think we need to update `app/models/user.rb` so that `current_user.name` returns the preferred name.

dkinzer: But on second thought, maybe it's too late to do that cause there are already a bunch of names in the DB.

sensei100: @dkinzer We don't want to get the current_user name.  We want to get the name from the Alma api now.

sensei100: Are you worried about backward compatibility?

dkinzer: Patly, But mostly I just don't like that we have to remember to look at both the controller and the model to get info/data about the user.  Seems like all of that should be in one place (preferably the model by convention).

For example `current_user.can_purchase_order?` is already getting information from alma.

sensei100: I guess I don't really think of that as an issue because I didn't make any changes to the controller.  I'm just calling @user, which is used in the rest of that template already.

dkinzer: Also we get all this other info from `current_user`, `current_user.loans`, `current_user.fines`, `current_user.holds` etc.

sensei100: Everything else used in that template is not defined in the model.

PR #1435, opened 2019-09-11
Bl 972 accessibility review Part 1

sensei100:
This one is getting big, so I am going to break it into parts.  This part focused on the catalog search results availability.  There are also some changes to the headers and navigation elements.

- The alma availability is currently rendered dt and dd tags that are not within a dl.  This updates the method to create a dl instead of a span.  It also renames the method to better explain what is being rendered.
- We have several places where duplicate ids are being used.  This adds the document_id to bookmarks hidden_fields and the online availability buttons to make the ids unique.
- We were missing some important landmarks.  This adds landmarks to the main content and sidebar.  It updates the footer role and adds unique labels to the navigation landmarks to better define them.
Fix syntax errors in search results
  - Adds empty alt tag to decorative icons
  - Removes comma in link 
  - Removes div inside of span
  - Closes span tag for electronic links
- BL-944 Add skip to main content link for keyboard navigation
- BL-934 Remove extra h2 element in search results

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -52,6 +52,10 @@ $(window).on('turbolinks:load', function() {
 
 $(document).ready(function() {
 
+	$('.decorative').each(function() {
+    $(this).attr('alt', "");
+  });
+
 	// This is necessary because iOS is triggering the resize event when an element is clicked.
 	// More information about this solution can be found here: https://stackoverflow.com/a/24212316/256854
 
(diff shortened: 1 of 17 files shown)


app/assets/javascripts/application.js
```diff
@@ -52,6 +52,10 @@ $(window).on('turbolinks:load', function() {
 
 $(document).ready(function() {
 
+	$('.decorative').each(function() {
```
dkinzer: @sensei100 what does this do?  Thanks.

sensei100: @dkinzer It adds an empty string alt attribute to elements with the decorative class so that an accessibility error is not thrown.  That code comes from the website navigation menu where the icons are added as background images.  You can't add an alt tag to a background image, so that has to be handled here instead.

PR #1467, opened 2019-09-20
Bl 973 refactor google analytics events

sensei100:
- Updates the code to use both ids and classes
- Removes duplicate ids from code
- Adds new listeners for features that didn't exist before
- Removes listeners for elements that have been removed
- Sets timeout to allow availability to be loaded first

### DIFF
--- a/app/assets/javascripts/listeners.js
+++ b/app/assets/javascripts/listeners.js
@@ -14,7 +14,7 @@ $(document).ready(function(){
 			{id: "navbar_journals", category: "navigation"},
 			{id: "navbar_website", category: "navigation"},
 			{id: "startOverLink", category: "search-results"},
-			{id: "start_over", category: "search-results"},
+			{id: "start_over", category: "record-page"},
 			{id: "back_to_search", category: "search-results"},
 			{id: "advanced-search-submit", category: "search-results"},
 			{id: "articles_advanced_search", category: "search-results"},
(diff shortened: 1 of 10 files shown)


app/assets/javascripts/listeners.js
```diff
@@ -1,66 +1,74 @@
...
-        eventLabel: event,
-        forceSSL: true,
-        anonymizeIp: true
-      });
-    }
-  }
+	  const handleEventClicks = (label, category) => {
+	    if (typeof ga != "undefined") {
+	      ga("send", "event", {
+	        eventCategory: "category",
```
dkinzer: This needs to be `category`
```suggestion
	        eventCategory: category,
```

app/assets/javascripts/listeners.js
```diff
@@ -1,66 +1,74 @@
...
+			if(track.id) {
+				if (el = document.getElementById(track.id)) {
+					el.addEventListener("click", () => {
+				    handleEventClicks(track.id, track.category)
+					});
+				};
+			} else {
+				let elements = [... document.getElementsByClassName(track.class)]
+				elements.forEach((el) => {
+					console.log(track.class)
```
dkinzer: Let's remove debugging code
```suggestion
```

app/assets/javascripts/listeners.js
sensei100: We were waiting for the okay from Emily before using categories.

PR #1496, opened 2019-09-30
BL-1011 Availability alerts

sensei100:
- We currently display an alert that availability can't be found for items with blank items_json_display fields.
- Electronic items don't have those fields, so the method is throwing an error.  Refactor the method to not include items with the electronic_resource_display field.

### DIFF
--- a/app/helpers/availability_helper.rb
+++ b/app/helpers/availability_helper.rb
@@ -62,9 +62,11 @@ module AvailabilityHelper
   end
 
   def availability_alert(document)
-    document["items_json_display"].map { |item|
-      item["availability"].blank?
-    }.any?
+    unless document["electronic_resource_display"]
+      document["items_json_display"].map { |item|
+        item["availability"].blank?
+      }.any?
+    end
   end
 
   def description(item)
--- a/app/helpers/availability_helper.rb
+++ b/app/helpers/availability_helper.rb
@@ -62,11 +62,10 @@ module AvailabilityHelper
   end
 
   def availability_alert(document)
-    unless document["electronic_resource_display"]
+    document["electronic_resource_display"].blank? &&
       document["items_json_display"].map { |item|
         item["availability"].blank?
       }.any?
-    end
   end
 
   def description(item)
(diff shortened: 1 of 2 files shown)


app/helpers/availability_helper.rb
```diff
@@ -62,9 +62,11 @@ def document_and_api_merged_results(document, items_list)
   end
 
   def availability_alert(document)
-    document["items_json_display"].map { |item|
-      item["availability"].blank?
-    }.any?
+    unless document["electronic_resource_display"]
```
dkinzer: @sensei100,  Thanks. This works, but I'd like to keep `availability_alert` to be a function that will return either `true` or `false`... 

Could we just add `document["items_json_display"].blank? &&`

spec/helpers/availability_helper_spec.rb
```diff
@@ -841,5 +841,30 @@
...
+          "electronic_resource_display" =>
+          [
+            { "title" => "Access electronic resource.", "url" => "http://libproxy.temple.edu/login?url=http://www.aspresolver.com/aspresolver.asp?SHM2;1772483" },
+            { "portfolio_id" => "77777", "title" => "Sample Name" },
+          ]
+          }
+        }
+
+      it "returns nil"  do
+        expect(availability_alert(document)).to eq nil
```
dkinzer: Once `availability_alert` returns true/false we can update this spec to check for `false` and make it consistent with test above this one where expectation is also `false`.

app/helpers/availability_helper.rb
sensei100: @dkinzer Can you be more specific in where the && would go?  I'm not sure  where to do that since right now we are mapping over a field that potentially doesn't exist.

dkinzer: Sure I think
```suggestion
    document["electronic_resource_display"] &&
```
will work.

spec/helpers/availability_helper_spec.rb
dkinzer:
```suggestion
        expect(availability_alert(document)).to eq false
```

PR #1537, opened 2019-10-22
Update ruby version

sensei100:


dkinzer:
We'll want to follow up and get qa ruby updated too.

PR #1538, opened 2019-10-22
[BL-953] Redirect back to request_options on login.

dkinzer:
This change updates the redirect process that happens when a user
authenticates right after trying to get request options for an item.

Instead of getting redirected to the search results page, the user is
redirected to the request options form.

This process will only be initiated the first time a user is redirected
to the request options form.  Once authenticated any other request will
proceed via the modal window as usual.

The reasons that I'm redirecting directly to the request form instead of
the results page and then opening the modal window are:
* It takes longer to wait for results page to load and then wait for
modal window to load
* The JavaScript to make that happen is rather clunky, and in my opinion
will be harder to maintain.

If this approach doesn't suffice we can try the JavaScript solution
which requires us to bind the modal trigger to the appearance of the
Availability button.  The availability button is not present initially
on the load page so we have to add an observer that looks for the
availability button to appear, then checks that it's the button that the
user initially clicked, then clicks it programatically.


sensei100:
@dkinzer I wanted to test this locally, but redirects don't work on my local.  Does the request options page that you redirect to have styles?

dkinzer:
@sensei100 yes, it renders with styles.

PR #1559, opened 2019-10-30
[BL-953] Open availability request modal on redirect.

dkinzer:


dkinzer:
Good to know.  I was in the process of turning the code into a module so that we can actually test it.  Yeah.. it should work, though.

dkinzer:
Actually, I mean BlacklightAlma but I guess come to think of it, I could also just turn the specific code into a module.  FYI... To test this locally when you are logged on You go to a URL like:

http://localhost:3000/catalog?utf8=%E2%9C%93&search_field=all_fields&q=and#doc-991036865390903

Note that the mms_id must be of the doc that actually shows up in the search page (that specific id works on my local but not necessarily on yours).

sensei100:
@dkinzer Are you going to make the changes to move the code to a different file?  If so, I will wait for that before testing.

PR #1574, opened 2019-11-05
[BL-1008] Fix web content search error.

dkinzer:
Fixes error that gets thrown when user does a web content search
immediately after an advanced search for a field not defined in web
content.

REF BL-1008
REF https://github.com/projectblacklight/blacklight_advanced_search/pull/106


dkinzer:
closing in favor of #1583

PR #1609, opened 2019-12-05
BL-985 Web content solrconfig

sensei100:
- Remove default_solr_params from web content controller


dkinzer:
I think we need to combine this with a switch to using the solrcloud collection which uses the correct configs in .env.qa/.env.stage/.env.prod

PR #1628, opened 2019-12-18
BL-1088 Blacklight modal bug

sensei100:
- Changing our jQuery to 3 in the application.js file caused the modal functionality to stop working. Revert back to requiring only jQuery

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -10,7 +10,7 @@
 // Read Sprockets README (https://github.com/rails/sprockets#sprockets-directives) for details
 // about supported directives.
 //
-//= require jquery3
+//= require jquery
 //= require 'blacklight_advanced_search'
 //= require rails-ujs
 //= require popper


dkinzer:
@sensei100 I just read this about the error we encountered `$(...).modal is not a function`: http://net-informations.com/jq/iq/modal.htm . It says that the error is do to the order of the javascript libraries.  I'd like us to look into that real quick before merging this because it would be nice to keep parity with what BL is doing upstream.

sensei100:
@dkinzer I don't see that as the problem.  jQuery is defined first and then the other dependencies. I don't know where we got that jQuery3 because when I look at the project Blacklight asset generator I see them using jQuery also.  Both on master and 7.4.2
Screen Shot 2019-12-18 at 8 52 42 AM

dkinzer:
OK, I'm just confused because `$('p').modal()` works on https://demo.projectblacklight.org . and I'm trying to figure out why that is (since they are using jquery-3).  What are we doing different?

sensei100:
That screenshot from the asset generator code is the same as ours.

dkinzer:
On my local when I make it exactly like what they have on the demo site then `$('p').modal()` works.   So I'm guessing it's one of the other libraries.

sensei100:
@dkinzer It is the blacklight-range-limit dependency.  If you remove that, it works.  But we definitely don't want to remove that, so I think for now we should commit this change and create an issue on the blacklight-range-limit repo.

dkinzer:
OK, the issue is coming from `blacklight_range_limit` (which is probably why they've recently upgraded it).  We can deal with it another time to get parity once we upgrade blacklight_range_limit.

PR #1638, opened 2019-12-19
Switch back az to non solrcloud (just in case PR)

dkinzer:


dkinzer:
Been long enough now that we can just close this.

PR #1647, opened 2020-01-06
BL-637 Move catalog solr conf to solrconfigs.xml.

dkinzer:
And,
* Update local docker.
* Set up correct configs for az, web and catalog content.
* Update circleci to use docker
* Update local and circleci to use solr8.


bibliotechy:
So, my reading is that the addition of the submodules is just a convenience for testing and local dev? 
Two questions I have are:
1) Any risk that code starts getting written that depends on their presence? 
2) Will you need to manage the submodule pointers to ensure the repos are pointing to the right config when running tests or local development? That seems like it could get complicated.

Not blocker concerns just wondering if you considered those issue in choosing submodules as opposed to a bash script to pull them in, etc.

dkinzer:
@bibliotechy yes, we would have to manage those.

I did  consider making the solr config repos  into a gems so that we'd get help managing.  I'm open to that, but it seems weird to do for configuration only/ non ruby repos.

On the other hand, submodules has a benefit that we can cd into submodule folders and push PRs to submodules repos directly from there.

dkinzer:
> Any risk that code starts getting written that depends on their presence?

I think this is very much mitigated by the fact that in our solrcloud set up we need to create tagged releases and we manage those using aliases.  So we're more likely to forget to create a release and update our solr URLs to point to the new release.  But that will be an issue regardless of wether we use submodules ore not and we would catch that on qa and stage.

ebtoner:
BL ticket reference should be BL-673, not BL-637.

PR #1670, opened 2020-01-20
BL-1048 Message on open shelving materials

sensei100:
- We currently display a banner message on all items that are in the MAIN library.  This is misleading for items that are not in the bookbot, but are also not on open shelving(reserves, reference, etc.)
- Refactors the code to display the message based on location and library.  The message should display for any items found in these locations: hirsch, juvenile, leisure, newbooks, stacks

### DIFF
--- a/app/helpers/availability_helper.rb
+++ b/app/helpers/availability_helper.rb
@@ -88,7 +88,7 @@ module AvailabilityHelper
     when "SCRC"
       render partial: "scrc_instructions", locals: { key: key, document: document }
     when "MAIN"
-      if main_stacks_message(key, document) == true
+      if main_stacks_message(key, document)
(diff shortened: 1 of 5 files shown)


app/helpers/availability_helper.rb
```diff
@@ -74,14 +74,23 @@ def availability_alert(document)
...
+
   def library_specific_instructions(key, document)
     case key
     when "ASRS"
       render partial: "asrs_instructions", locals: { key: key }
     when "SCRC"
       render partial: "scrc_instructions", locals: { key: key, document: document }
     when "MAIN"
-      render partial: "main_open_shelving_instructions", locals: { key: key }
+      if main_stacks_message(key, document) == true
```
dkinzer:
```suggestion
      if main_stacks_message(key, document)
```

PR #1675, opened 2020-01-30
Allow getting a list of collections via Alma::Electronic::Collections API.

dkinzer:


sensei100:
@dkinzer Should this be closed now that you did the work in a different way?

dkinzer:
Yup, this is no longer good.

PR #1679, opened 2020-02-03
Update qa solr to use new SC config.

dkinzer:
Also fixes controller not all fields queries.

And, updates local to point latest tul_cob-catalog-solr.

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -253,31 +253,32 @@ class CatalogController < ApplicationController
     # case for a BL "search field", which is really a dismax aggregate
     # of Solr search fields.
 
+
     config.add_search_field("title") do |field|
       # solr_parameters hash are sent to Solr as ordinary url query params.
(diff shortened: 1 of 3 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -253,31 +253,32 @@ class CatalogController < ApplicationController
     # case for a BL "search field", which is really a dismax aggregate
     # of Solr search fields.
 
+
     config.add_search_field("title") do |field|
       # solr_parameters hash are sent to Solr as ordinary url query params.
-      field.solr_parameters = { 'spellcheck.dictionary': "title" }
```
sensei100: @dkinzer Did you mean to remove the spellcheck.dictionary? We kept it for author and subject, so I'm just double checking.

dkinzer: @sensei100 thanks for catching that.  I noticed also that the rest were actually being overridden since now they are all `field.solr_parameters` so I fixed that too.

PR #1733, opened 2020-03-18
BL-1150 adding banner

sensei100:
- This adds a banner to the application due to the temporary closure


sensei100:
@ebtoner Updated again.  The circleci issue is what I referenced on the Blacklight channel this morning.  There is a problem with our configs that are causing all PRs to fail at the moment.  We are working on fixing it.

PR #1740, opened 2020-03-20
BL-1152 Updates e-service notes.

dkinzer:
Also updates e-service notes script to always get latest set of electronic collection ids.

REF: BL-1152


bin/alma_electronic_db.rb
```diff
@@ -9,18 +9,13 @@
...
-filename = "spec/fixtures/electronic-collections-ids.csv"
-
-ids = File.readlines(filename)
-  .map(&:to_i).select(&:nonzero?)
+ids =  Alma::Electronic.get_ids
   .map { |id| { collection_id: id.to_s } }
 
 batch = Alma::Electronic::BatchUtils.new(ids: ids)
 
 batch.get_collection_notes
```
sensei100: @dkinzer Are we still using these methods?

dkinzer: yes

PR #1794, opened 2020-04-06
Switch button content for mobile using browser gem

bibliotechy:
Adds use of browser alias to application so it is picked up at start up.

Adds some media query fanciness for button padding on mobile.


bibliotechy:
@sensei100 thanks for catching that.

PR #1800, opened 2020-04-09
Don't show button for archival materials

bibliotechy:
Fix a slight button drrop on small screens.


sensei100:
@bibliotechy Rubocop offense for trailing whitespace.

PR #1833, opened 2020-04-22
Move e-notes to db.

dkinzer:
We are running into a weird caching issue with e-notes in Rails.configuration.


app/helpers/catalog_helper.rb
```diff
@@ -449,6 +455,7 @@ def render_electronic_notes(field)
     collection_notes = get_collection_notes(collection_id)
     service_notes = get_service_notes(service_id)
     unavailable_notes = get_unavailable_notes(service_id)
+    require "pry"; binding.pry
```
sensei100, line 458:
```suggestion
   
```
We should remove this

db/schema.rb
```diff
@@ -10,7 +10,28 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2018_01_09_214210) do
+ActiveRecord::Schema.define(version: 2020_04_22_020027) do
+
+  create_table "active_storage_attachments", force: :cascade do |t|
```
bibliotechy, line 15: Why is this active storage stuff here?

db/migrate/20200422020027_create_electronic_notes_bags.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class CreateElectronicNotesBags < ActiveRecord::Migration[5.2]
+  def change
+    create_table :electronic_notes_bags do |t|
+      t.string :note_type, index: { unique: true }
+      t.json :value
```
bibliotechy, line 7: I feel like timestamps would be useful in case we want to check when this instance was created

db/schema.rb
dkinzer: This keeps getting added to my local build ... I'm not sure why.  I pushed it up in case it's required by something.

db/migrate/20200422020027_create_electronic_notes_bags.rb
dkinzer: Good point .. I'll add them back

db/schema.rb
dkinzer: @bibliotechy ok, I tried removing it.. but as soon as I re ran rake db:migrate it comes back.

bibliotechy: @dkinzer So, the migrations that would create the updates in schema are not in your db/migrate, so my guess is that you ran those migrations at some point locally. 

The need migrations were included as part of Rails 5.2 upgrade by default, but y'all probably removed them since we had no use case for ActiveStorage in this app. 

`db/schema.rb` is generated from your actual local database. So, I'd advise you to drop your data base, recreate it and then run the migrations in this branch again and they should be removed.

dkinzer: OK, thanks.  I'm trying that now.

dkinzer: @bibliotechy Yup, after I dropped my local db and re ran the migration those tables are no longer added.  Thanks!

PR #1842, opened 2020-04-28
[WiP] Move all the docker files to .docker folder

dkinzer:
It's easier to know that all docker files are in one subfolder vs some in root and some in subfolder.


dkinzer:
Still needs some dots and i's cross... will come back to it later

dkinzer:
not worth the effort

PR #1845, opened 2020-04-30
Set up next release.

dkinzer:


sensei100:
@dkinzer This doesn't look right.  All of the hathitrust updates have already been deployed, but they are showing up again.

dkinzer:
@sensei100 this is because either squashed merged those to qa the last time or because it's a previous cherry pick hotfix is cofusig git.. But there's not much we can do about this.

PR #1856, opened 2020-05-12
Update tul_cob CI to be master branch only.

dkinzer:
CI will now.
* Deploy to tul_cob/qa on merges to master branch.
* Automatically deploy to stage when a new release is created.
* Deploy to production on confirmation when a new release is created.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,14 +11,15 @@ workflows:
          filters:
            branches:
              only:
-               - qa
+               - master
   stage_deploy:
     jobs:
       - stage_deploy:
           filters:
             branches:
-              only:
-                - master
+              ignore: /.*/
+            tags:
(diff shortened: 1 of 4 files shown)


bibliotechy:
@dkinzer I think now would probably be a good time to explicitly document what the process for getting things onto QA, stage and master. We have https://github.com/tulibraries/tul_cob/wiki/Create-a-new-tagged-release-for-tul_cob which will become out of date. Something like the [release docs we have in manifold](https://github.com/tulibraries/manifold/blob/qa/docs/cutting-releases.md)

dkinzer:
@bibliotechy OK, I've pushed up some documentation.

dkinzer:
If we decide to do this, we'll still need to update the stage deploy script ... holding any more work on this until after we have a group discussion.

dkinzer:
I've updated the stage deploy script.  Should be ready to go now.

PR #1857, opened 2020-05-12
BL-1186. Update referer in chat link.

dkinzer:
Use https://librarysearch.temple.edu as the referer.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,10 +46,10 @@ jobs:
 
       - restore_cache:
           keys:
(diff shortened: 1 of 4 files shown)


dkinzer:
@sensei100 I'm not sure, I think there needs to be parity between the referrer coming from header and the one from sidebard widget.  Also that header referer is set to localhost which I think is definitely bad since that's a developer thing.

@cdoyle-temple can you verify that both the referrer coming from chat that is in navbar or chat coming from sidebar should both be "https://librarysearch.temple.edu", please.

PR #1885, opened 2020-05-29
Update qa build to use master branch.

dkinzer:
* Use --limit=app instead of --skip=solr so that it's more like production.
* Use the master branch instead of default set by playbook.


dkinzer:
Yeah, we just thought this would be faster.  We're cool either way.

bibliotechy:
@dkinzer I understand that it will work this way, but I'm not comfortable leaving that inconsistency in place. The[ playbook README](https://github.com/tulibraries/tul_cob_playbook#deployments), for example, indicates that CD handles the deployment for Prod but not for QA. In like a month we're gonna forget this is configured in this place and spend like an hour wondering how it even works. I'd like to see a PR for the playbook as well before merging this

PR #1899, opened 2020-06-04
BL-1201 Update airflow catalog production solr collction.

dkinzer:
* On deployments to production make required airflow changes.

- [x] Depends on tulibraries/ansible-playbook-airflow#190


.circleci/config.yml
```diff
@@ -144,3 +144,7 @@ jobs:
             - "59:d3:45:4c:e5:31:4e:cd:64:c0:e3:2a:73:a3:55:94"
       - run:
           command: bash .circleci/deploy-prod.sh
+
+      - run:
+          name: Update Airflow On Deployment
```
bibliotechy, line 149: This could be a little more specific, something like
```suggestion
          name: Set Production Airflow Variable "CATALOG_PRODUCTION_SOLR_COLLECTION" to match value in .env
```

PR #1904, opened 2020-06-08
Bl 1187 chat link changes

quoideneuf:


ebtoner:
@quoideneuf @dkinzer I think the referrer should still be librarysearch.temple.edu, not library.temple.edu?

PR #1923, opened 2020-06-11
allow user to use searchCDI param to toggle proquest view

quoideneuf:
adds a hack to meet item 3 in the BL-1213 requirements. Maybe there is a blacklightier way to do this...?


dkinzer:
@quoideneuf Just curious if this can be done in the search builder and be kept close to the the other code instead of separating it out.

ebtoner:
Adding a quick clarification here, too, since I'm not sure that it was fully clear in the Jira issue -- the proquest id is just an example, and this update should apply to non-proquest id's as well. Happy to provide additional info, if needed.

quoideneuf:
🦀
     🦀 
           🦀 
           🌊

PR #1939, opened 2020-06-18
quick fix strip off trailing backslashes

quoideneuf:
This should prevent issues like this for the time being:

https://app.honeybadger.io/projects/56250/faults/64881431


quoideneuf:
well, this won't work. probably better to leave as is until it can be fixed correctly - the error is only triggered by an unusual search string.

PR #1944, opened 2020-06-23
BL-1171: Update blacklight_range_limit.

dkinzer:
* Updates blacklight_range_limit to latest version.
* Patches JS issue (projectblacklight/blacklight_range_limit#111)
* Hides remove button (to keep current behavior)
* Patches render_constraints_filters method (projectblacklight/blacklight_range_limit#152)

Does not override the range limit view because we would need to override the entire view and thus lose possible upgrade changes.

New version:
image

Old version:
image

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -56,7 +56,7 @@ gem "tzinfo-data", platforms: [:mingw, :mswin, :x64_mingw, :jruby]
 gem "blacklight", "~> 7.8.0"
 gem "blacklight_advanced_search", git: "https://github.com/projectblacklight/blacklight_advanced_search.git"
 gem "blacklight-marc", git: "https://github.com/projectblacklight/blacklight-marc.git", ref: "v7.0.0.rc1"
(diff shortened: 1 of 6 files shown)


amecynth:
David Kinzer, I'm looking at the new date slider on QA. I am OK with the slider being on the top instead of the bottom of the date range. I narrowed my date range once and the layout was fine. But then I narrowed it for a second time and the height of the date range tripled.
Screen Shot 2020-06-24 at 8 30 47 AM

dkinzer:
@amecynth thanks for reporting. I believe @ebtoner has mentioned this issue in the past to me but it's hard to reproduce consistently.

Note however I have a patch for that issue in this PR which should do away with it. But it has not yet bee merged and not ready to qa/testing just yet.

amecynth:
Ok, thanks! In the scope of all the work that needs to be done, this making this display completely perfect is a lower priority, so I would say it looks good and go ahead and implement whatever patch you have.

PR #1955, opened 2020-07-01
Devo 109 deprecation audit

quoideneuf:
* Also seems to fix BL-1254


quoideneuf:
Setting this to ready for review. I will continue to scan for any breakage this causes. Maybe during the next maint sprint we can continue to remove helper methods and view template logic where possible and move to presenters and components.

app/helpers/blacklight_helper.rb
```diff
@@ -1,30 +0,0 @@
-# frozen_string_literal: true
```
dkinzer, line 1: This looks good to me except for one error I'm running into when article bookmark is not found.

In order to reproduce the issue create a new `PrimoCentralDocument`  Bookmark via the rails console with a bogus id that you know will fail. (ex. `Bookmark.new(user_id: 6, user_type: "User", document_type: "PrimoCenralDocument", document_id: "xyzfoobarxyz")` [user would be different in your case]).  Then go look at your Bookmarks in the browser.

When I do this I get the following error
```
Showing /usr/local/bundle/gems/blacklight-7.10.0/app/views/catalog/_index_header.html.erb where line #18 raised:

undefined method `presenter' for nil:NilClass
```
I can resolve the error locally by doing `git checkout main -- app/helpers/blacklight_helper.rb`

Since we're sunsetting article bookmarks I'm not going to spend too much time (adding a test for instance to this case, however, I think we need to not break current functionality as some users would have bookmarked articles which are no longer valid and this PR would break their bookmarks pages.

PR #1991, opened 2020-07-15
#BL-358 Display due date in availability

nomadicoder:
Addresses Issue BL-358

- Expand Alma query to get due_date from holding
- Parse date string to display in common mm/dd/yyyy format (Jira Issues specifies m/d/yyyy)

### DIFF
--- a/app/helpers/availability_helper.rb
+++ b/app/helpers/availability_helper.rb
@@ -36,6 +36,13 @@ module AvailabilityHelper
   def unavailable_items(item)
     if item.has_process_type?
       process_type = Rails.configuration.process_types[item.process_type] || "Checked out or currently unavailable"
+      if (item.process_type == "LOAN")
+        due_date_time = item["item_data"].fetch("due_date", nil)
+        unless (due_date_time.nil?)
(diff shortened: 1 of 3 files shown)


app/helpers/availability_helper.rb
```diff
@@ -36,6 +36,13 @@ def availability_status(item)
   def unavailable_items(item)
     if item.has_process_type?
       process_type = Rails.configuration.process_types[item.process_type] || "Checked out or currently unavailable"
+      if (item.process_type == "LOAN")
+        due_date_time = item["item_data"].fetch("due_date", nil)
+        unless (due_date_time.nil?)
+          due_date = due_date_time.match(/([[:digit:]]+)-([[:digit:]]+)-([[:digit:]]+)/)
+          process_type += sprintf(", due %i/%i/%i", due_date[2].to_i, due_date[3].to_i, due_date[1].to_i)
+        end
```
dkinzer, line 44: Please refactor to use the [`make_date`](https://github.com/tulibraries/tul_cob/blob/main/app/helpers/users_helper.rb#L10) method so that we can maintain consistency with loan date feature and also take timezone into account.

PR #1994, opened 2020-07-16
[BL-1245] Feature flag enable recall link.

dkinzer:
* When we're back on campus, enable recall link for unavailable items.

- [x] Depends on #1999


dkinzer:
@bibliotechy I'm cool with that ... but because of the way you implemented your helper method as a controller method that gets pushed to helper... there's no easy way to test that.  Would you mind if I move your CatalongController.campus_closed? into helpers?

PR #2007, opened 2020-07-20
Only show Find On Shelves Message if an item available

bibliotechy:


app/views/almaws/request_options.html.erb
```diff
@@ -8,7 +8,9 @@
   </div>
 
   <div class="panel-group request-group" id="request-accordian-<%= @mms_id %>">
-  <%= open_shelves_partial(@request_options, @books, @document) %>
+  <% if @items.any?(&:in_place?) %>
+    <%= open_shelves_partial(@request_options, @books, @document) %>
+  <% end %>
```
dkinzer, line 13: I have an old backlog issue where I've been wanting to move and use document items data into our document model.  But it keeps getting pushed back.  It's partially done.  There is a [`document.materials`](https://github.com/tulibraries/tul_cob/blob/v1.2.5/app/models/solr_document.rb#L93-L103) that gets items for the document. And that can be used by the `open_shelves_partial` helper to do what you are doing here where it could be more easily tested and maintained over time.

bibliotechy: @dkinzer I really like this idea, but yeah look slike it is not quite ready for primetime. 

I just tried this in pry, and it triggered an additional API request.

```
From: /Users/cbn/projects/tul_cob/app/helpers/catalog_helper.rb @ line 207 CatalogHelper#open_shelves_allowed?:

    204: def open_shelves_allowed?(document)
    205:   relevant_locations = ["hirsh", "juvenile", "leisure", "stacks", "newbooks"]
    206:
 => 207:   binding.pry
    208:   document.fetch("items_json_display", []).any? { |item|
    209:     item["current_library"].include?("MAIN") } &&
    210:   document.fetch("items_json_display", []).any? { |item|
    211:     relevant_locations.include?(item["current_location"]) }
    212: end

[1] pry(#<#<Class:0x00007f915c1bc750>>)> document.materials
{"type":"alma_bib_item","mms_id":"991037387908903811","limit":100,"timestamp":"2020-07-20 11:47:10 -0400","referer":"http://localhost:3000/catalog/991037387908903811","user_agent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:78.0) Gecko/20100101 Firefox/78.0","duration":3.683438}
=> [{"title"=>"The ocean reader : history, culture, politics /", "barcode"=>"23494288290003811", "call_number"=>"GC21 .O266 2020", "library"=>"Charles Library", "location"=>nil, "availability"=>"Checked out"}]
```
Yeah, I was actually hoping for something just like you describe, but not finding it, went with my solution.

Big 👍 to that backlog issue

PR #2008, opened 2020-07-20
ensure that multi-valued fields on show page are rendered as links

quoideneuf:


app/views/catalog/_show_fields.html.erb
```diff
@@ -5,7 +5,7 @@
 <% else %>
   <dd class="blacklight-<%= field_name.parameterize %> col-xs-12 col-sm-9">
     <ul>
-    <% field_presenter.values.each do |value| %>
+    <% Array.wrap(field_presenter.render).each do |value| %>
```
dkinzer, line 8: @nomadicoder won't this basically always render something even where there are no fields to render?  `Array.wrap(nil)` will render stuff.

quoideneuf: @dkinzer - good catch, added some tests and reworked the template a bit.

quoideneuf:
this needs more work - I'll reopen shortly

quoideneuf:
I can't seem to reopen this...

PR #2032, opened 2020-07-29
force all primo requests to use CDI

quoideneuf:
If you are merging this, something bad has happened.


quoideneuf:
Don't think we'll ever need this, but if we do we can always reopen it.

PR #2043, opened 2020-08-03
Bl 1165 sorting by call number

quoideneuf:


app/controllers/concerns/facet_params_dedupe.rb
```diff
@@ -0,0 +1,15 @@
...
+  extend ActiveSupport::Concern
+
+  included do
+    # Converts facet params like ?f[foo]=bar&f[foo]=bar to ?f[foo]=bar
+    before_action do
+      if params["f"]
+        facet_params = params.extract!("f")
+        params.merge!(f: facet_params["f"].transform_values { |v| v.is_a?(Array) ? v.uniq : v })
+      end
+    end
```
dkinzer, line 13: hmmm... @quoideneuf I'm curious about how does this scenario come about?  Are you safeguarding against users adding facets params directly into URL?  Or are we generating URLs with multiple params somewhere (like via redirects or something)?

quoideneuf: @dkinzer This is a contrived example, where I have added a repetition directly to the address bar by cutting and pasting:
image

This one I did just by clicking. I managed to double the "English" facet by selecting it in the standard search, then clicking advanced search, then selecting it in again in the dropdown.

image

I implemented it this way just because I figured it would remove the possibility of parameter doubling sneaking back in as we go.

dkinzer: @quoideneuf if I'm understanding you correctly we are adding this code to guard against the user adding duplicate facet params in the URL themselves (not that it comes out naturally in some other context)?

quoideneuf: @dkinzer - it can also happen naturally, if one adds facets in basic search, jumps to advanced search, and selects the same facet from the drop down.

PR #2046, opened 2020-08-03
Ensure min range date for primo

quoideneuf:
there is probably more work to do here, but this stops a couple 500 errors from primo

### DIFF
--- a/app/models/blacklight/primo_central/search_builder_behavior.rb
+++ b/app/models/blacklight/primo_central/search_builder_behavior.rb
@@ -122,6 +122,7 @@ module Blacklight::PrimoCentral
 
       min = params.dig("range", "creationdate", "begin")
       max = params.dig("range", "creationdate", "end")
+      min = 1000 if min.to_s.match(/^\d+$/) && min.to_i < 1000
       range = YearRange.new(min, max)
       primo_central_parameters[:range] = range
 
--- a/app/models/blacklight/primo_central/search_builder_behavior.rb
+++ b/app/models/blacklight/primo_central/search_builder_behavior.rb
@@ -122,11 +122,12 @@ module Blacklight::PrimoCentral
 
       min = params.dig("range", "creationdate", "begin")
       max = params.dig("range", "creationdate", "end")
-      min = 1000 if min.to_s.match(/^\d+$/) && min.to_i < 1000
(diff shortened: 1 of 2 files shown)


app/models/blacklight/primo_central/search_builder_behavior.rb
```diff
@@ -122,6 +122,10 @@ def process_date_range_query(primo_central_parameters)
 
       min = params.dig("range", "creationdate", "begin")
       max = params.dig("range", "creationdate", "end")
+      min = 1000 if min.to_s.match(/^\d+$/) && min.to_i < 1000
+      if max.to_s.match(/^\d+$/) and max.to_i < min.to_i
+        max = min
```
dkinzer, line 127:
```suggestion
        (min, max) = [max, min]
```

I think this needs to swap the values, otherwise you will have effectively set both to the same value always.

Also maybe a test case for this context ... i.e. check that values get swapped?

app/models/blacklight/primo_central/search_builder_behavior.rb
```diff
@@ -122,7 +122,12 @@ def process_date_range_query(primo_central_parameters)
...
       min = params.dig("range", "creationdate", "begin")
       max = params.dig("range", "creationdate", "end")
-      range = YearRange.new(min, max)
+
+      min = (min.blank? || (min.to_s.match(/^\d+$/) && min.to_i >= 1000)) ? min : 1000
+      max = (max.blank? || (max.to_s.match(/^\d+$/) && max.to_i >= 1000)) ? max : 1000
+
+      min_max = min.blank? || max.blank? || min.to_i < max.to_i ? [min, max] : [max, min]
+      range = YearRange.new(*min_max)
       primo_central_parameters[:range] = range
```
dkinzer, line 131: Hmm.., to me this seems more complex than necessary or like we're handling the same problem in different places.  * Are we safe to just fix issue in Primo?
* If not, should we move this into YearRange?

@sensei100 , @bibliotechy  what do you think?

quoideneuf: I agree - I did this just with the aim of removing the possibility of creating an application error by putting bad values into the form. But perhaps it would be good to think about how we want to handle bad values from a usability perspective.

sensei100:
@dkinzer What do you want to do with this one?  Close or merge?

PR #2049, opened 2020-08-04
move general controller rescuing to a concern, handle primo errors

quoideneuf:
Improves error handling by:

- ensuring we don't send the rails 500 message to the end user
- consolidating error handling in a 'concern'
- giving us a headstart when we find out a primo query has caused an error.

See also:
https://github.com/tulibraries/primo/pull/35

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -8,6 +8,7 @@ class CatalogController < ApplicationController
 
   include BlacklightAlma::Availability
   include Blacklight::Marc::Catalog
+  include ServerErrors
 
   before_action :authenticate_purchase_order!, only: [ :purchase_order, :purchase_order_action ]
   before_action :set_thread_request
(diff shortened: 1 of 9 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -20,23 +21,6 @@ class CatalogController < ApplicationController
...
-    redirect_back(fallback_location: root_path, notice: "The start year must be before the end year.")
-  end
-
-  rescue_from Blacklight::Exceptions::RecordNotFound,
-    with: :invalid_document_id_error
-
-  rescue_from Blacklight::Exceptions::InvalidRequest do |exception|
-    Honeybadger.notify(exception.message)
-    render "errors/unsupported_query"
-  end
```
dkinzer, line 33: Why are we getting rid of these?

quoideneuf: @dkinzer I moved them into the server_errors concern. I thought it would be easier to keep everything in one place, and also make it convenient when developing to just comment out `include ServerErrors` in the controller.

quoideneuf: I guess that is fairly subjective, so happy to move them back if others prefer.

dkinzer: But aren't some specific errors to the user actually useful.  Like knowing a record was not found vs a backend networking issue?

dkinzer: Oh, I see what's going on... 😎

dkinzer: Argh.. sorry.. so I thought that your solution was still letting the user know for instance that record just doesn't exists 404... I think the very least we should do that one. @sensei100 @bibliotechy what do you think?

sensei100: As long as we are still going to get the specific error messages, I don't have a preference as to where they are located.

quoideneuf: @dkinzer I didn't intentionally change record not found behavior. Let me look into it a bit more. I did want to do some more testing of various scenarios anyway.

quoideneuf: @dkinzer @sensei100 I added a test for Primo::Search::ArticleNotFound error. The behavior should be the same, though I did set the status to 404 in the handler.

dkinzer:
Looking at this now.

PR #2070, opened 2020-08-19
BL-1294 alma holds that don't have expirations

quoideneuf:


quoideneuf:
For the interim release: 
https://github.com/tulibraries/tul_cob/pull/new/v1.3.3-cherry-pic

spec/controllers/users_controller_spec.rb
```diff
@@ -32,10 +32,38 @@
       end
 
       describe "GET #holds" do
+        render_views
+
         xit "returns http success" do
```
sensei100, line 37: @quoideneuf @dkinzer @bibliotechy I know this isn't part of these code changes, but in looking at this file I see we are skipping a number of tests.  It looks like that code was last changed 4 years ago.  Should we look at using these tests or remove them if they are not actually needed?

dkinzer: @sensei100 Yeah, I think that might be one of those default tests that gets added by spec template?  I'm OK with any of those getting removed.

quoideneuf: @dkinzer @sensei100 I can make a pass through these while I'm working on other error handling issues

PR #2076, opened 2020-08-25
BL-1079 Prevent parselet errors on empty advanced search fields

quoideneuf:
https://app.honeybadger.io/projects/56250/faults/53917673

### DIFF
--- a/app/models/search_builder.rb
+++ b/app/models/search_builder.rb
@@ -96,7 +96,7 @@ class SearchBuilder < Blacklight::SearchBuilder
   end
 
   def process_is(value, op)
-    return if value.blank?
+    return if value.nil?
     return value if value.match(/"/)
 
     if op == "is"
--- a/app/models/search_builder.rb
+++ b/app/models/search_builder.rb
@@ -96,8 +96,8 @@ class SearchBuilder < Blacklight::SearchBuilder
   end
 
   def process_is(value, op)
-    return nil if value.nil?
-    return value if value.match(/"/) rescue true
+    return if value.blank?
+    return value if value.match(/"/)
 
     if op == "is"
       "\"#{value}\""
--- a/app/models/search_builder.rb
+++ b/app/models/search_builder.rb
@@ -96,6 +96,7 @@ class SearchBuilder < Blacklight::SearchBuilder
   end
 
   def process_is(value, op)
+    return nil if value.nil?
     return value if value.match(/"/) rescue true
 
     if op == "is"
--- a/app/models/search_builder.rb
+++ b/app/models/search_builder.rb
@@ -97,8 +97,6 @@ class SearchBuilder < Blacklight::SearchBuilder
 
   def process_is(value, op)
     return if value.nil?
-    return value if value.match(/"/)
-
     if op == "is"
       "\"#{value}\""
     else
--- a/app/models/search_builder.rb
+++ b/app/models/search_builder.rb
@@ -97,6 +97,7 @@ class SearchBuilder < Blacklight::SearchBuilder
 
   def process_is(value, op)
     return if value.nil?
+    return value if value.match(/"/) rescue true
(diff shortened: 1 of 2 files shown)


app/models/search_builder.rb
```diff
@@ -96,6 +96,7 @@ def process_begins_with(value, op)
   end
 
   def process_is(value, op)
+    return nil if value.nil?
     return value if value.match(/"/) rescue true
 
     if op == "is"
```
dkinzer, line 102: @quoideneuf I'm confused by this change `return value if value.match(/"/) rescue true` should already return `nil` if value is `nil`.  And we have a unit test to prove it.  Is the issue that `q` is not `nil` but a blank like `""`?

@see https://github.com/tulibraries/tul_cob/blob/main/spec/models/search_builder_spec.rb#L150-L153

app/models/search_builder.rb
```diff
@@ -96,6 +96,7 @@ def process_begins_with(value, op)
   end
 
   def process_is(value, op)
+    return nil if value.nil?
     return value if value.match(/"/) rescue true
```
dkinzer, line 100: Maybe this should be
```suggestion
    return if value.blank?
    return value if value.match(/"/)
```

dkinzer: If this turns out to be the issue, then we should probably add a unit test for when `value=""` to `process_is` spec.

app/models/search_builder.rb
quoideneuf: @dkinzer I did include a controller test which fails without this line and passes with it. I didn't go much deeper into it than that.

app/models/search_builder.rb
quoideneuf: I will see if the tests pass with this version too,. Seems like it would.

app/models/search_builder.rb
```diff
@@ -96,7 +96,8 @@ def process_begins_with(value, op)
   end
 
   def process_is(value, op)
-    return value if value.match(/"/) rescue true
+    return if value.blank?
```
dkinzer, line 99: Arg it sees not to want to return nil if value is ""...

```suggestion
    return if value.nil?
```

I'm really confused by this error.. it's not happening at this method just farther down the stack (which is part of my confusion)

app/models/search_builder.rb
```diff
@@ -96,8 +96,7 @@ def process_begins_with(value, op)
   end
 
   def process_is(value, op)
-    return value if value.match(/"/) rescue true
```
dkinzer, line 99: @quoideneuf the problem with processing if value contains a quote (") is that multiple quotes can get added to value on multiple runs.  Maybe we should just keep both lines?  Though I still don't quite understand why the orional version wasn't returning nil when value was nil.

quoideneuf: @dkinzer I think maybe sometimes `value` is not a string? Two tests broke when the line in question was added: https://app.circleci.com/pipelines/github/tulibraries/tul_cob/1714/workflows/e4b97715-4704-41b9-bbfa-de253c40488b/jobs/2448

app/models/search_builder.rb
quoideneuf: @dkinzer ah, I missed that this run failed too. Let me have another look.

app/models/search_builder.rb
quoideneuf: @dkinzer ok, I got confused. I think in the original I did keep this line, then as this thread went on I forgot that and removed it. I think the problem with `value=nil` in the orginal is: it NoMethodErrors on `value.match`, then the line rescues to true. But it doesn't return true, it proceeds, and thus when `op="is"` it returns `\"\"`.

dkinzer: Yeah that makes sense.

PR #2079, opened 2020-08-26
DEVO-70: Do load fixtures instead of ingest.

dkinzer:


sensei100:
@dkinzer Why would we want to make this change?  Loading the fixtures is only going to give me a very small amount of data to work with, coming from the spec/fixture files.  The way it is currently written gives a variety of records to work with.

dkinzer:
Okay let me think about it some more. The reason to do it is because that will load a to z and web content data but maybe we should do that plus load the ingest stuff... I don't know I'll think about it

sensei100:
@dkinzer I would vote for adding az and web content to the ingest command.  I often rely on having a lot of catalog records when testing out features and would prefer for that to be the default.

dkinzer:
@sensei100 how about the other way round?  Saying because we are already using load_fixtures in the CI environment and it will line up nicely with my desire to keep that environment and our local as closely matched.  Also I've been wanting to get rid of one of those commands, because I'm finding it confusing to know how the CI environment and the local are getting stuff.

In fact I just now as I was trying to work out this PR I was having a hard time figuring out what was doing what.

dkinzer:
Hmm... On second thought, I think ingest by itself may still have some use ..

dkinzer:
OK, @sensei100 I've updated the code to now load sample data as well in the local context when `Make load_fixtures` is invoked.

And, I fixed `rake ingest` so that it still works as before, but now does not commit per file when loading all files in sample data.

dkinzer:
For some reason the sample data is now getting ingested in the CI environment even though I'm pretty sure it shouldn't be ... not sure if it's some kid of caching thing going on but I'm looking into it.

dkinzer:
Yup, issue was that `DO_INGEST` environment variable was being set to an empty string.

lib/tasks/tulsearch.rake
```diff
@@ -49,11 +58,16 @@ task :ingest, [:filepath] => [:environment] do |t, args|
   if file && file.match?(/databases.json/)
     az_url = Blacklight::Configuration.new.connection_config[:az_url]
     `SOLR_AZ_URL=#{az_url} cob_az_index ingest --use-fixtures --delete`
+  elsif file && file.match(/\/web_content/)
+    web_url = Blacklight::Configuration.new.connection_config[:web_content_url]
+    `SOLR_WEB_URL=#{web_url} cob_web_index ingest --use-fixtures --delete`
```
dkinzer, line 63: oops .. I just noticed that I'm using fixtures here.. should be passing file (idea is we added a file to web_content fixtures we want to ingest... arg.

PR #2089, opened 2020-09-01
BL-1307 Lib Guides endpoint

sensei100:
- Sets up basic route, controller, and model for Lib Guides API

### DIFF
--- a/app/models/lib_guides_api.rb
+++ b/app/models/lib_guides_api.rb
@@ -1,0 +1,61 @@
+# frozen_string_literal: true
+
+class LibGuidesApi
+  attr_reader :query
+
+  class << self
+    alias fetch new
+  end
+
+  def initialize(query)
+    @query = query
+  end
+
+  def base_url
+    config["base_url"]
+  end
+
+  def api_key
+    config["api_key"]
+  end
+
+  def config
+    Rails.configuration.lib_guides
+  end
+
+  def as_json(*)
+    guides
+  end
+
+  private
+
+    def guides
+      json.take(3)
+    end
+
+    def json
+      JSON.parse(response)
+    rescue JSON::ParserError => e
+      Honeybadger.notify("Parsing LibGuides JSON response at #{url} failed with #{e}")
+      []
+    end
+
+    def response
+      @response ||= begin
+        http = HTTParty.get(url)
+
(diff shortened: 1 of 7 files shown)


app/models/lib_guides_api.rb
```diff
@@ -0,0 +1,61 @@
...
+      @response ||= begin
+        http = HTTParty.get(url)
+
+        if http.success?
+          http.body
+        else
+          "[]"
+        end
+      end
+    rescue HTTParty::ConnectionFailed => e
```
quoideneuf, line 53: I don't find this in the HTTParty namespace:
https://www.rubydoc.info/github/jnunemaker/httparty/toplevel

Perhaps just rescue all Exceptions here, and respond with status 504 and empty json? It would probably be good for the ajax client to know if there was an error.

app/models/lib_guides_api.rb
```diff
@@ -0,0 +1,61 @@
...
+  def base_url
+    config["base_url"]
+  end
+
+  def api_key
+    config["api_key"]
+  end
+
+  def config
+    Rails.configuration.lib_guides
```
quoideneuf, line 23:
```suggestion
    Rails.application.config_for(:lib_guides)
```

quoideneuf: (Don't know if I am missing some Rails knowledge here, but this didn't work for me as it is. )

dkinzer: This line should work as long as we configure it the way we do our other stuff, like here: https://github.com/tulibraries/tul_cob/blob/main/config/application.rb#L24-L36

app/models/lib_guides_api.rb
dkinzer: There's also `HTTParty::Error` as an alternative

app/models/lib_guides_api.rb
sensei100: @quoideneuf I updated the application.rb file to include that config.  That way we remain consistent with previous work.

app/models/lib_guides_api.rb
sensei100: @quoideneuf Good catch.  I updated it to rescue all exceptions.

PR #2090, opened 2020-09-01
Update lib_guides_controller.rb

quoideneuf:


sensei100:
@quoideneuf Thanks for the catch.  I'm not sure if you already know this, but you can make suggestions for small fixes like this instead of creating a whole new PR.

PR #2091, opened 2020-09-01
BL-1297: Add export article bookmarks button.

dkinzer:
Add temporary feature add allow users to export their bookmarks.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -36,6 +36,9 @@ en:
           <p>
             <a href="https://library.temple.edu/contact-us">Contact us</a> if you have questions or any issues accessing/exporting your article bookmarks.
           </p>
+          <p>
+            Please note that only Item URLs will be exported. You can use Item URL to locate the item in the catalog and download the bibliographic information
+          </p>
 
     borrow_html: "<li>If you can't find something, %{href}.</li>"
     borrow_href: "borrow from another library"
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -37,7 +37,7 @@ en:
             <a href="https://library.temple.edu/contact-us">Contact us</a> if you have questions or any issues accessing/exporting your article bookmarks.
           </p>
           <p>
-            Please note that only Item URLs will be exported. You can use Item URL to locate the item in the catalog and download the bibliographic information
+            Please note that only Item URLs will be exported. You can use Item URL to locate the item in the catalog and download the bibliographic information.
           </p>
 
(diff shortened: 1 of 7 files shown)


config/locales/en.yml
```diff
@@ -36,6 +36,9 @@ en:
           <p>
             <a href="https://library.temple.edu/contact-us">Contact us</a> if you have questions or any issues accessing/exporting your article bookmarks.
           </p>
+          <p>
+            Please note that only Item URLs will be exported. You can use Item URL to locate the item in the catalog and download the bibliographic information
```
sensei100, line 40:
```suggestion
            Please note that only Item URLs will be exported. You can use Item URL to locate the item in the catalog and download the bibliographic information.
```

spec/controllers/bookmarks_controller_spec.rb
```diff
@@ -101,4 +101,31 @@ class FooDocument < SolrDocument; end
...
+      get :export_articles
+      # We successfully export test file
+      expect(response).to be_successful
+      expect(response.content_type).to eq("text/plain")
+
+      description = response["Content-Disposition"]
+      expect(description).to eq("attachment; filename=\"article_bookmark_urls.txt\"")
+
+      # We export list of URLs
+      urls = <<~EOT
```
sensei100, line 119: @dkinzer I really want to go back to the discussion of using this.  I don't understand why that is necessary for a small amount of text. I understand the use case more when you are copying longer xml or json data, but this is only two lines.

dkinzer: @sensei100 In this case I don't have any justification other than it looks better to me.  Note that I did take your advice from before and indented lines like you wanted by using `<<~` instead of `<<-`.

Also note that in this case from a non stylistic point of view it my make sense to use a here document format as we are literally generating a document on the fly that we are exporting to the user.

config/locales/en.yml
dkinzer: Thanks

spec/helpers/bookmark_helper_spec.rb
```diff
@@ -37,5 +37,34 @@
...
+          expect(helper.render_article_bookmark_export_button(docs)).to be_nil
+        end
+      end
+
+      context "article documents present" do
+        doc_a = SolrDocument.new(id: "foo")
+        doc_b = PrimoCentralDocument.new({})
+        docs = [doc_a, doc_b]
+
+        it "returns nil" do
```
dkinzer, line 62:
```suggestion
        it "returns an html button wrapped in list item tag" do
```

PR #2094, opened 2020-09-02
BL-1308 Build url for libguides

sensei100:


app/models/lib_guides_api.rb
```diff
@@ -56,6 +56,17 @@ def response
...
-      # TBD
+      query_terms = {
+        site_id: site_id,
+        key: api_key,
+        sort_by: "relevance",
+        search_terms: "#{query}"
+      }
+
+      URI::HTTPS.build(
+        host: "example.com",
```
dkinzer, line 67: @sensei100 is this still in draft?  Otherwise `host: "example.com"` seems wrong.

sensei100: @dkinzer You are right.  I updated it now.

PR #2101, opened 2020-09-04
Fix styling bug

sensei100:
- The automatic row styles were causing extra spacing to appear on small screen sizes.  Adding no-gutters eliminates the problem.

### DIFF
--- a/app/views/catalog/_results_pagination.html.erb
+++ b/app/views/catalog/_results_pagination.html.erb
@@ -1,0 +1,9 @@
+<% if show_pagination? and @response.total_pages > 1 %>
+ <div class="row record-padding no-gutters">
+  <div class="col-md-12">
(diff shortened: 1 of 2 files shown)


dkinzer:
@sensei100 this PR seems to also be adding conditional pagination to a page?

sensei100:
@dkinzer No, this is over-riding a blacklight template.

dkinzer:
@sensei100 cool, in that case could we make a note on the template that we're overriding this from blacklight.  (Not sure if we've been doing this for templates, but I find it very useful sometimes when I'm hutting down a bug to know wether issue is do to local code or whether I need to also look upstream for bug).

sensei100:
@dkinzer I added a comment

dkinzer (approved):

PR #2102, opened 2020-09-04
Do not swallow the test errors.

dkinzer:
Errors are being swallowed locally when running tests (Rails env == test) and this is making it harder to figure out what is going on.

### DIFF
--- a/app/controllers/concerns/server_errors.rb
+++ b/app/controllers/concerns/server_errors.rb
@@ -4,7 +4,7 @@ module ServerErrors
   extend ActiveSupport::Concern
 
   included do
-    next if Rails.env == "development"
+    next if ["development", "test"].include?(Rails.env)
 
     # We handle all unknown and unplanned exceptions here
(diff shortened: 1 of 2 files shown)


quoideneuf:
@dkinzer I'm not sure I understand this. It makes sense to divert error handling for development, but don't you sometimes want to test that/how an error is handled?

dkinzer:
@quoideneuf yes, we will still test  how this works in the test environment by temporarily switching Rails.env to "production". Without this change errors get slowed up and it's and it's not entirely obvious that you aren't seeing the whole error until some head scratching. 

@quoideneuf note that the qa and stage environments use Rails.env == "production" So we will continue to see user behavior on those environments.

Note that Rails.env == "test" only gets set by running `rspec`  or `rake ci` (which is not what is happening on qa/stage/production)

dkinzer:
arg... looks like I have a couple more places to update Rails.env temporarily during the test.

PR #2110, opened 2020-09-10
BL-1102 Request bug

sensei100:
- We are currently throwing errors when a user selects a material type and a description that does not exist for the selected material type.
- This code groups the descriptions by material types and filters the description dropdown based on the material type selected

### DIFF
--- a/app/lib/cob_alma/requests.rb
+++ b/app/lib/cob_alma/requests.rb
@@ -175,7 +175,7 @@ module CobAlma
     def self.physical_material_type_and_descriptions(items_list)
       types_and_descriptions = items_list.all.map { |item|
         Hash[item.physical_material_type["desc"], [item.description]] unless item.physical_material_type["value"] == ""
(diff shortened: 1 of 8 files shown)


app/lib/cob_alma/requests.rb
```diff
@@ -161,6 +172,24 @@ def self.physical_material_type(items_list)
       material_types.uniq.compact
     end
 
+    def self.physical_material_type_and_descriptions(items_list)
+      types_and_descriptions = items_list.all.map { |item|
+        Hash[item.physical_material_type["desc"], [item.description]] unless item.physical_material_type["value"] == ""
+      }.uniq
```
dkinzer, line 178:
```suggestion
      }.uniq.compact
```

@sensei100 without `.compact` at the end this can produce `[nil]` which will cause an error below

quoideneuf (approved):

PR #2111, opened 2020-09-10
BL-1306 javascript for libguides recommender

quoideneuf:
client side controller for loading lib guide data. the librarian section is missing some fields that exist in the mockups but not in the api. 

Still needs error handling, tests, etc.

### DIFF
--- a/app/assets/stylesheets/partials/_bento.scss
+++ b/app/assets/stylesheets/partials/_bento.scss
@@ -9,15 +9,15 @@
 		column-count: 2;
 	}
 	@include media-breakpoint-only(lg) {
-		column-count: 2;
+		column-count: 3;
 	}
 	@include media-breakpoint-only(xl) {
-		column-count: 2;
+		column-count: 3;
 	}
 }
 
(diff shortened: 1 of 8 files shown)


sensei100:
@quoideneuf This looks good, but I think we may need to look at using javascript to inject the lib guides section into the search results.  It is supposed to display after the first 3 results, but it is displaying at the end.  If you are planning on implementing that as part of the styling ticket, I can merge this one.  I think we also need clarification from Emily/Jackie about whether it should display when there are no catalog search results.

quoideneuf:
> @quoideneuf This looks good, but I think we may need to look at using javascript to inject the lib guides section into the search results. It is supposed to display after the first 3 results, but it is displaying at the end. If you are planning on implementing that as part of the styling ticket, I can merge this one. I think we also need clarification from Emily/Jackie about whether it should display when there are no catalog search results.

@sensei100 I didn't realize that, but can do it as part of the styling.

sensei100 (approved):

PR #2120, opened 2020-09-14
Refactor to only show lib-guides in the catalog search results

sensei100:


spec/features/indices_spec.rb
```diff
@@ -132,5 +132,17 @@
...
       end
     end
+
+    feature "Lib Guides section" do
+      scenario "Bookmarks page" do
+        visit "/bookmarks"
+        expect(page).to_not have_css(".lib-guides-recommender-catalog")
+      end
+
+      scenario "Bookmarks page" do
```
quoideneuf, line 142:
```suggestion
      scenario "Catalog page" do
```

quoideneuf (approved):

PR #2122, opened 2020-09-14
BL-1052: Map location facet constraints

dkinzer:
Use library location label maps in constraints section.

### DIFF
--- a/app/helpers/facets_helper.rb
+++ b/app/helpers/facets_helper.rb
@@ -43,15 +43,40 @@ module FacetsHelper
     @locations_map ||= Rails.configuration.locations.values.inject(&:merge)
   end
 
+  def library_location_label(value, include_library = false)
+    library, label = value.split(" - ")
+    label = locations_map[label] || label
+
+    if include_library
+      [library, label].join(" - ")
+    else
+      label
+    end
+  end
+
   def pre_process_library_facet!(item)
     # Filter out secondary facets that do not match library
     item.items.select! { |i| i.value.match?(/#{item.value}/) }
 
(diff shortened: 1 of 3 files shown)


dkinzer:
@ebtoner here is what this would look like currently
image

Does that work for you?

ebtoner:
@dkinzer Thanks for pinging me on this. There are two components here: 
- Mapping the location code: This was my request in the original Jira comment. The mapping itself looks fine (it's now displaying BookBot rather than ASRS), though it should also take into account the comments in the next bullet point.
- Relationship between the library and location values: There are other changes in this PR that I did not request. The facet value should have the combined library and location values, like it currently does on QA -- ex. “Library > Charles Library - Bookbot”. This makes it specific to that location within the library rather than a generic location. 

Other possible updates: On QA currently, the location facet also selects a second facet value for higher-level library -- ex. "Library > Charles Library." We could eliminate that when a more specific location is selected, so that only the library/location combo displays. This is what Jackie has asked Brian to do for call numbers. 

If you have any questions, feel free to talk it over with Chad and Jackie.

dkinzer:
@ebtoner, thanks; @quoideneuf has a PR opened that deals how pivot constraints will get rendered which should take care of the the second point #2127

dkinzer:
So I tested that @quoideneuf PR against this PR and it doesn't look like it addresses the concerns entirely so I'll try to fix that.

dkinzer:
@bibliotechy I've updated to address @ebtoner second point

image

bibliotechy (approved):

bibliotechy (approved):

PR #2127, opened 2020-09-15
BL-1164-nested-facets-ui-fixes

quoideneuf:
Modifications to the pivot facet logic:

- Renders a nesting facet as unselected if it contains a selected nested facet. The hyperlink of the nesting facet should effectively remove the nested facet and add the nesting facet if it isn't already there.
- Renders nesting facet in open state if it contains a selected nested facet.

Tests are still needed, but perhaps it would be better to wait until it has been evaluated for usability.

### DIFF
--- a/app/helpers/facets_helper.rb
+++ b/app/helpers/facets_helper.rb
@@ -4,7 +4,7 @@ module FacetsHelper
   include Blacklight::FacetsHelperBehavior
 
   def facet_item_component_class(facet_config)
-    return super if facet_config.pivot
+    return FacetItemPivotComponent if facet_config.pivot
     default_component = FacetItemComponent
     facet_config.fetch(:item_component, default_component)
   end
@@ -54,7 +54,6 @@ module FacetsHelper
     }
   end
 
-
   ##
   # Overridden to allow pivot sub fields to be rendered in 'selected' state
   # and pre process the "library_pivot_facet" field.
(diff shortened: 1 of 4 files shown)


dkinzer (changes requested):
This looks great! But for some reason tests are failing.  I'll try to move that long if I can figure it out with a PR against this one.  Also, I think this code is complicated enough that some tests would be useful for future maintenance/development and just general understanding of how the code is supposed to be working.

I'll see if I can add some of that too if I can get to it.

dkinzer:
So I tested further but with the change the locations_facets change and I think it ends up braking that somewhat ... once we make one selection we can't make any other
image


Also for some reason I thought this PR would handle constraints but, it is not touching that at all.

bibliotechy:
@dkinzer I'm not seeing the behavior you are talking about, or I'm misunderstanding your comment.  See below, I've selected BookBot but also have the option to select more locations
Screen Shot 2020-09-16 at 10 30 16 AM

dkinzer:
@bibliotechy that's not behavior I'm seeing when I merge #2122 to this PR but I removed my previous comment as I believe it matches closer to what @ebtoner actually desires. I will work on the Other PR to address constraint representation issue @ebtoner brought up in that #2122 PR which is not yet been addressed.

dkinzer:
ok, I'm my local I just had some other facets enabled 😊

dkinzer (approved):

dkinzer:
Added followup issue: https://tulibdev.atlassian.net/browse/BL-1325

app/helpers/facets_helper.rb
```diff
@@ -69,10 +68,33 @@ def render_facet_item(facet_field, item)
...
+      presenter.hide_facet_param(args[:hide_facet_param])
+      presenter.keep_in_params!
+    end
+    args.delete(:hide_facet_param)
+    facet_item_component_class(facet_config).new(facet_item: presenter, **args).with_view_context(self)
+  end
+
+  def facet_item_presenter(facet_config, facet_item, facet_field)
+    FacetItemPresenter.new(facet_item, facet_config, self, facet_field)
+  end
```
dkinzer, line 99: @quoideneuf I hadn't realized this was an override when I merged this.  It threw me for a loop when I was working on my last PR because I was trying to pass in a specific presenter via the configuration.

I think we need to add some method documentation explaining that we have overridden this function and why why.

I'm assuming that adding our presenter via configuration was not working?

quoideneuf: @dkinzer Responding to your last question above, unfortunately, this one is hardcoded in blacklight for some reason:
https://github.com/projectblacklight/blacklight/blob/master/app/helpers/blacklight/facets_helper_behavior.rb#L315-L317

PR #2130, opened 2020-09-16
BL-676 Set up background jobs

sensei100:
- Sets up the use of background jobs for sending texts and emails
- This is working locally running sidekiq and a redis server


bibliotechy:
@sensei100 this looks great, but I don't want to merge because we don't have redis available in any environment.

We should either:

1. switch the queue adapted to `:inline` and then merge so we can test that the logic works as expected
2. move this into draft until playbook changes are made to add redis support.

Not sure if you have a preference.

PR #2138, opened 2020-09-18
avoid using SearchState::to_h when tweaking facet href params

quoideneuf:
Blacklight::SearchState::to_h returns the actual param hash, so using it to render facet params causes bad side effects (deleting search params). So we can just utilize the existing remove and add functions, which return deep copies of that facet hash.

### DIFF
--- a/app/presenters/facet_item_presenter.rb
+++ b/app/presenters/facet_item_presenter.rb
@@ -23,16 +23,12 @@ class FacetItemPresenter < Blacklight::FacetItemPresenter
     @keep_in_params
   end
 
-  def href(path_options = {})
-    if selected? && keep_in_params?
-      search_path(search_state.to_h)
-    else
-      super
-    end
-  end
-
   def remove_href(path = search_state)
-    search_path(path.remove_facet_params(facet_config.key, facet_item))
+    if keep_in_params?
+      search_path(path.remove_facet_params(nil, nil))
+    else
+      search_path(path.remove_facet_params(facet_config.key, facet_item))
+    end
   end
 
   def add_href(path_options = {})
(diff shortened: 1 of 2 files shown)


app/presenters/facet_item_presenter.rb
```diff
@@ -23,18 +23,14 @@ def keep_in_params?
     @keep_in_params
   end
 
-  def href(path_options = {})
-    if selected? && keep_in_params?
-      search_path(search_state.to_h)
+  def remove_href(path = search_state)
+    if keep_in_params?
+      search_path(path.remove_facet_params(nil, nil))
```
dkinzer, line 28: I think we should rename `path` `search_state` so that there's no ambiguity as to what that is.

Also I'm confused about the line `path.remove_facet_params(nil, nil)` do we mean to pass (nil, nil) or is that a mistake?

quoideneuf: yes, it is intentional. we just want a clean deep copy of the params, and while `to_h` does not do that, `remove_facet_params` does. This is for cases where we drilled down and are building the link for the parent facet - it is selected, but unlike a normally selected facet, we do want it to be applied if we decide to move back up in the hierarchy.

quoideneuf: as for the names, I just kept the signature the same as the base method:
https://github.com/projectblacklight/blacklight/blob/master/app/presenters/blacklight/facet_item_presenter.rb#L58-L60
happy to rename path, but search_state is built in. rename `path` to `_search_state` maybe?

dkinzer: @quoideneuf thanks for the explainer.  I guess I'm OK with the keeping name as is if it's the same as the signature.  Seems an odd choice to me.  In other parts of the code where they use the similar signature they do `local_search_state = search_state`. ...like herehttps://github.com/projectblacklight/blacklight/blob/master/app/helpers/blacklight/render_constraints_helper_behavior.rb#L29

I guess since it's a private function they're more relax about naming 🤷

dkinzer (approved):

PR #2140, opened 2020-09-18
Update libguides search terms and response handling for better relevance

bibliotechy:
This pull requests updates the way we generate search queries to send to the libguides API and changs how we handle the API response before generating html.

Instead of sending only the original search term to the libguides API, we now grab the orginal search term plus the top 10 subject_facet_topic terms, wrap each term (which can be multiple words) in parenthesis and then join all of the terms with a boolean `OR`. This allows for a much broader search than the possibly very narrow original search query.

Additionally, we also process the response from the libguides API to boost Subject Guides and Topic Guides over Course Guides and General Purpose guides.


quoideneuf (approved):

dkinzer (approved):
Looks great!

dkinzer:
🤖 👮‍♂️

PR #2142, opened 2020-09-19
BL-1052: Remove location translation step.

dkinzer:
- [x] Depends on tulibraries/cob_index#141

REQUIRES FULL_REINDEX

Move location translation step to indexing time because doing that
handles collapsing items that should be treated as coming from the same
location.  Otherwise they appear as duplicates or have the wrong facet
count.


dkinzer:
Converted to draft, because I don't want to merge this until we do a Full Reindex

bibliotechy (approved):

PR #2154, opened 2020-09-23
Use cob_index locations map to avoid duplication.

dkinzer:


bibliotechy:
👍

sensei100 (commented):
@dkinzer What do we want to do about the locations.rake task then? It doesn't make sense to kept here if we aren't using that file anymore.

dkinzer:
Good point @sensei100. Let's move that to cob_index.

dkinzer:
Although, hmmm.. it's tricker because right now cob_index is blissfully unaware of any alma configs I think.

bibliotechy:
@dkinzer cob_index knows about Alma configs...though maybe not in the most re-usable way https://github.com/tulibraries/cob_index/blob/main/lib/cob_index.rb#L40-L44

dkinzer:
Oh, in that case, 😎

bibliotechy:
@dkinzer I'd say that once https://github.com/tulibraries/cob_index/pull/143 is merged, update this to remove that rake task and we're good to go!

dkinzer:
@bibliotechy, done.  But I can't merge because it's still blocked until it gets at least one official approval.

bibliotechy (approved):

PR #2155, opened 2020-09-23
Fixes a couple deprecation warnings

sensei100:
- We were overriding a file that was causing a problem in Blacklight version 7.4.  David submitted a fix to BL that was merged, so that file is being removed.
- Updates the way we call default_per_page method
- Fixes typo in helper file


dkinzer (changes requested):
@sensei100 blacklight_helper override is still needed until we either remove article bookmarks altogether or they fix this issue: https://github.com/projectblacklight/blacklight/issues/2234

PR #2160, opened 2020-09-24
pivot facet styling, and refactor facet item component and presenter

quoideneuf:


app/helpers/facets_helper.rb
```diff
@@ -50,46 +50,6 @@ def pre_process_library_facet!(item)
...
   end
 
-  ##
-  # Overridden to allow pivot sub fields to be rendered in 'selected' state
-  # and pre process the "library_pivot_facet" field.
-  #
-  def render_facet_item(facet_field, item)
-    if facet_field == "library_pivot_facet"
-      pre_process_library_facet!(item)
-    end
```
dkinzer, line 60: The facet values are no longer getting pre_processed ... Should be `BookBot` not `Charlse Library - BookBot`... but I did notice that the selection I do might not be necessary.

image

app/helpers/render_constraints_helper.rb
```diff
@@ -8,9 +8,19 @@ def filter_elements(facet, values, search_state)
...
       presenter = facet_item_presenter(facet_config, val, facet)
 
+      if facet == "lc_inner_facet" && search_state.filter_params["lc_outer_facet"]
+        label = "#{search_state.filter_params['lc_outer_facet'][0]} | #{label}"
+        presenter.parent = OpenStruct.new(field: "lc_outer_facet", value: search_state.filter_params["lc_outer_facet"][0])
+      end
+
+      if facet == "location_facet" && search_state.filter_params["library_facet"]
+        presenter.parent = OpenStruct.new(field: "library_facet", value: search_state.filter_params["library_facet"][0])
+      end
```
dkinzer, line 22: I'm confused as to why this is needed?

dkinzer (changes requested):

dkinzer:
One other thing I noticed is that at least on my local it's not rendering the way it did in your presentation for locations.  Or it could be that still requires an lc facet to be enabled to render.

app/helpers/facets_helper.rb
quoideneuf: @dkinzer I moved that into the presenter, take a look and let me know if it looks ok:
https://github.com/tulibraries/tul_cob/pull/2160/commits/1939e2b58d50d42a4eb3cbd51eb79f4a28432363

app/helpers/render_constraints_helper.rb
quoideneuf: @dkinzer This is preventing behavior where, if you click to cancel the restraint "Bookbot", the restraint for "Charles Library" will take its place. The presenter tacks the "parent" on when it is building the "remove_href" value. Because we are in the constraint pipeline though, we don't have actual objects for "parent" facets like we do when rendering facets.erb. But the struct looks enough like a FacetItem object for the presenter to build the link.

dkinzer: Maybe we should add a comment about that in the code.  It's not obvious.

quoideneuf:
> One other thing I noticed is that at least on my local it's not rendering the way it did in your presentation for locations. Or it could be that still requires an lc facet to be enabled to render.

@dkinzer I'm a little stumped by this. Is it possibly a caching issue? The styles should be the same:
https://github.com/tulibraries/tul_cob/pull/2160/files#diff-d2a8cf130312fdf5d4f441c694caa596R4

app/controllers/catalog_controller.rb
```diff
@@ -149,8 +149,7 @@ def self.with_call_number_facet?
     config.add_facet_field "genre_facet", label: "Genre", limit: true, show: true, component: true
     config.add_facet_field "genre_full_facet", label: "Genre", limit: true, show: false, component: true
     config.add_facet_field "language_facet", label: "Language", limit: true, show: true, component: true
-    config.add_facet_field "lc_facet", label: "LC Call Number", pivot: ["lc_outer_facet", "lc_inner_facet"], limit: true, show: true, component: true, collapsing: true, if: with_call_number_facet?
-
+    config.add_facet_field "lc_facet", label: "LC Call Number", pivot: ["lc_outer_facet", "lc_inner_facet"], limit: true, show: true, component: true, collapsing: true, icons: { show: "", hide: "" }, if: with_call_number_facet?
```
bibliotechy, line 152:
```suggestion
    config.add_facet_field "lc_facet", label: "Library of Congress Classification", pivot: ["lc_outer_facet", "lc_inner_facet"], limit: true, show: true, component: true, collapsing: true, icons: { show: "", hide: "" }, if: with_call_number_facet?
```

bibliotechy:
@dkinzer @bh3 I pulled this down locally and both LC and Library facets are rendering as expected, even when LC facets are not present.

dkinzer:
@bibliotechy It's not rendering for me as expected so I'm not sure what's going on.

bibliotechy:
@bh3 OK, a quick chat with @dkinzer and I see his point. And actually, I am seeing the extra X even when an LC facet is slected
Screen Shot 2020-09-25 at 11 36 03 AM

quoideneuf:
@bibliotechy that doesn't look right. I am afraid that something is wrong with git, but I am not sure how to resolve. I may need to take this branch down and re-post.

quoideneuf:
@bibliotechy @dkinzer Would you mind pulling down this branch and seeing if it makes a difference? 

https://github.com/tulibraries/tul_cob/tree/BL-1131-facet-styling-squashed

dkinzer:
@quoideneuf, it's not making a difference for me. Perhaps issue is something local is not yet committed?

bibliotechy:
@quoideneuf @dkinzer same. No difference.

app/presenters/facet_item_presenter.rb
```diff
@@ -1,48 +1,54 @@
...
-  # specific facet and replace it with the more general facet.
-  # So we don't want the specific one sticking around.
-  def hide_facet_param(facet_item)
-    @hidden_facet_params ||= []
-    @hidden_facet_params << facet_item
+  def items
+    items = super
+    if facet_field == "library_pivot_facet"
+      items = items.select { |i| i.value.match?(/#{facet_value}/) }
+      items.each { |i| i.label = i.value.split(" - ", 2).reject { |j| j.blank? }.last }
```
dkinzer, line 8: Why don't we use the helper method that I wrote here where we can less intrusively update at will?  Also I think at least we should add the comments back, because it's not obvious what or why we're doing this.

I don't understand the reject bit as the original code was working without that

app/components/facet_item_pivot_component.rb
```diff
@@ -3,31 +3,51 @@
...
+  # so that we can show the selected facet in the side panel
   def uncollapse?
-    return view_context.facet_field_in_params?(@facet_item.items[0].field) &&
-           params["f"][@facet_item.items[0].field] &&
-           params["f"][@facet_item.items[0].field].include?(@facet_item.items[0].value)
+    return false unless params["f"]
+    if @facet_item.facet_item.field == "library_facet"
+      return @facet_item.items.any? { |i|
+        params["f"][i.field] && params["f"][i.field].include?(i.value) && i.value.match?(/#{@facet_item.facet_item.value}/)
+      }
```
dkinzer, line 13: This step might be redundant now that you are basically doing the same thing in `def items` override.

spec/helpers/facets_helper_spec.rb
```diff
@@ -43,60 +43,4 @@ def self.name
...
-
-    context "sub item has 'foo - bar' value" do
-      let(:sub_item_a) { Blacklight::Solr::Response::Facets::FacetItem.new(value: "foo - bar", hits: 5, items: []) }
-      let(:item) { Blacklight::Solr::Response::Facets::FacetItem.new(value: "foo", hits: 5, items: [ sub_item_a ]) }
-
-      it "transforms label to just be 'bar'" do
-        label = item.items.first.label
-        expect(label).to eq("bar")
-      end
-    end
```
dkinzer, line 90: We should add these tests back or add the equivalent in the function we update to do the same thing

app/presenters/facet_item_presenter.rb
```diff
@@ -1,48 +1,56 @@
...
-  def hide_facet_param(facet_item)
-    @hidden_facet_params ||= []
-    @hidden_facet_params << facet_item
+  def items
+    items = super
+    if facet_field == "library_pivot_facet"
+      # Filter out secondary facets that do not match library
+      items = items.select { |i| i.value.match?(/#{facet_value}/) }
+      # Add proper secondary facet labels
+      items.each { |i| i.label = i.value.split(" - ", 2).reject { |j| j.blank? }.last }
```
dkinzer, line 10: Is the `.reject { |j| j.blank? }` step necessary?  On qa all location labels are properly rendered without this.

app/presenters/facet_item_presenter.rb
quoideneuf: there was a value in my fixtures that turned up with a blank after the dash. But I can't see it on prod so I'll take the reject out.

quoideneuf:
@bibliotechy @dkinzer Ok, there was a glitch that I wasn't seeing locally due to my smaller data set. There should only be 1 'X' rendered now when you select an inner facet.

app/presenters/facet_item_presenter.rb
quoideneuf: Took this out - there was a record that had a "Library - " value but in my local collection, but I don't find any on production.

app/components/facet_item_pivot_component.rb
quoideneuf: 👍 Took out the whole thing in fact - it was all redundant.

app/helpers/render_constraints_helper.rb
quoideneuf: done

app/helpers/facets_helper.rb
quoideneuf: should be fixed now.

dkinzer (approved):
This is working nicely for me now.  Assuming tests pass we should move forward with merge.

PR #2165, opened 2020-09-25
Fix advanced search library faceting.

dkinzer:
Only brings to parity with current broken solution.


bibliotechy:
@dkinzer I'm a little nervous about knock-on effects on removing the inclusive facet param, as that seems to be pretty central to a lot of what blacklight_advanced_search is doing, but I don't see any implications here, so merging

bibliotechy (approved):

PR #2166, opened 2020-09-25
fix broken advanced search pivot facet fields

quoideneuf:
@dkinzer This is an alternative to:
https://github.com/tulibraries/tul_cob/pull/2165

https://github.com/tulibraries/tul_cob/pull/2165 might be more straightforward if it doesn't have any bad side effects.


bibliotechy:
Closed in favor of #2165

PR #2190, opened 2020-10-12
Remove unnecessary configuration.

dkinzer:
* The relative URL root is set to the default ("/") for all our
environments. Therefore this configuration is no longer needed.


quoideneuf (approved):

bibliotechy:
@dkinzer We'll want to do something similar in the playbook 
https://github.com/tulibraries/tul_cob_playbook/blob/qa/group_vars/all.yml#L30

PR #2191, opened 2020-10-12
Allow primo and alma apikey config using environment variable.

dkinzer:
We already do this for Alma requests.  We should do it for Primo
requests too.  In K8 we will be setting secret configs via environment
variables.

### DIFF
--- a/config/initializers/alma.rb
+++ b/config/initializers/alma.rb
@@ -1,12 +1,13 @@
 # frozen_string_literal: true
 
-Alma.configure do |config|
-  # You have to set te apikey
-  config.apikey = Rails.configuration.alma[:apikey]
-  config.enable_loggable = true
-  config.timeout = Rails.configuration.alma[:timeout] || 30
-end
 ENV["ALMA_API_KEY"] ||= Rails.configuration.alma[:apikey]
 ENV["ALMA_DELIVERY_DOMAIN"] ||= Rails.configuration.alma[:delivery_domain]
 ENV["ALMA_INSTITUTION_CODE"] ||= Rails.configuration.alma[:institution_code]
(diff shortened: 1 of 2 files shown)


bibliotechy:
@dkinzer Would it make more sense to do this in [config/bento.yml](https://github.com/tulibraries/tul_cob/blob/main/config/bento.yml.example) where a bunch of other variables for Primo are listed? Or even migrate what is in bento.yml to `config.primo.yml`?

dkinzer:
Hmm.. @bibliotechy if we do it that way it means we'll have to manually remove config/bento.yml (note this file it's not committed. That link points to bento.yml.example).

I had thought we were already doing this for Alma [here](https://github.com/tulibraries/tul_cob/blob/main/config/initializers/alma.rb#L9).  But it looks like we are actually not.  Imho I think we should go further there and also use ENV variables as the ultimate overriding feature.

bibliotechy:
Regarding `bento.yml`, if we replace https://github.com/tulibraries/tul_cob/blob/main/config/application.rb#L30 with a line pointing to a primo config file, then it doesn't matter if that bento.yml file is still on the sever, not in version control. We stop loading it and start loading a new file and then we can go clean it up at our leisure. 

The new file can use the better approach of looking up from env.

dkinzer:
@bibliotechy Oh, OK.  I get what you mean.

dkinzer:
hmm...  @bibliotechy on second thought I'm still confused... we can't commit the apikey to git.  So even if we have a new file we still have to point to a new file that we add manually.  I wanted to do this without doing a dependent PR on tul_cob_playbook because I just need it to play around with k8 stuff.  I think I'll go ahead an put this on draft for now until we get that dependent tul_cob_playbook out first.

dkinzer:
Closing this because I intent to make it part of another related PR on k8 config work.

PR #2201, opened 2020-10-15
Setup some k8 specific docker configs.

dkinzer:


.dockerignore
```diff
@@ -1,3 +1,21 @@
...
 tags
+
+node_modules
+!/log/.keep
+!/tmp/.keep
+
+.gitignore
+
+.git
+.gitignore
```
sensei100, line 11: @dkinzer This is already on line 8
```suggestion

```

.dockerignore
```diff
@@ -1,3 +1,21 @@
...
+.DS_Store
+README.md
+/Dockerfile
+/log
+
+## Needed on CirlceCI
+/solr
+/sample_data
```
sensei100, line 19: @dkinzer Doesn't CircleCI just need the spec/fixtures?  I think we only use sample data to populate our locals.

dkinzer: It needs solr and spec ... but all these can removed from the production environment.

sensei100: @dkinzer But I think that /sample data can be moved up and not included with the Needed on CircleCI group.  Circle doesn't use any of those files.

dkinzer: oh, i see what you mean... everything in this file is getting ignored... the comment is just for us to keep track of stuff that might brake ci while we work through this.  I think we wont need the comment anymore.

.dockerignore
```diff
@@ -1,3 +1,63 @@
...
+
+
+.git
+.gitignore
+.DS_Store
+README.md
+/Dockerfile
+/log
+
+## Needed on CirlceCI
```
sensei100, line 16:
```suggestion

```

sensei100: Can we remove the comment then?

config/environments/production.rb
```diff
@@ -60,7 +60,7 @@
   config.log_tags = [ :request_id ]
 
   # Use a different cache store in production.
-  config.cache_store = :mem_cache_store
+  #config.cache_store = :mem_cache_store
```
sensei100, line 63: What happens here if we don't use the men_cache_store?

config/initializers/alma.rb
```diff
@@ -1,12 +1,13 @@
...
 
+ENV["ALMA_API_KEY"] ||= Rails.configuration.alma[:apikey]
+ENV["ALMA_DELIVERY_DOMAIN"] ||= Rails.configuration.alma[:delivery_domain]
+ENV["ALMA_INSTITUTION_CODE"] ||= Rails.configuration.alma[:institution_code]
+ENV["ALMA_AUTH_SECRET"] ||= Rails.configuration.alma[:auth_secret]
+
 Alma.configure do |config|
   # You have to set te apikey
-  config.apikey = Rails.configuration.alma[:apikey]
+  config.apikey = ENV["ALMA_API_KEY"]
```
sensei100, line 10:
```suggestion
  config.apikey = ENV["ALMA_API_KEY"] ||= Rails.configuration.alma[:apikey]
```

The primo file still has the config as backup.  Should we do that here too?

config/environments/production.rb
dkinzer: We don't have memcached set up in k8 yet so we need to comment this out for now

config/initializers/alma.rb
dkinzer: @sensei100 we are already doing that on line 3 above.  So we do not need to redo it again.

config/environments/production.rb
dkinzer: hmmm... actually I realize now we'll need to uncomment this because it will affect production hmmm....

dkinzer: OK, updated change so that it will not affect our regular production configuration just the K8 deployments.

config/initializers/alma.rb
quoideneuf: Ok for me to commit this? I think it is more readable and prevents an inattentive reorder breaking things.

config/initializers/alma.rb
```diff
@@ -1,12 +1,13 @@
 # frozen_string_literal: true
 
+ENV["ALMA_API_KEY"] ||= Rails.configuration.alma[:apikey]
```
quoideneuf, line 3:
```suggestion
```

quoideneuf (approved):

dkinzer:
reconverted to draft to think through some things based on discussion

config/initializers/alma.rb
dkinzer: I reverted changes to this fie.

PR #2210, opened 2020-10-21
BL-1323 Do not hit alma on database page.

dkinzer:
Skips adding stimulus controller to document show page when viewing
databases as databases have not alma availability information that needs
to be loaded.


sensei100:
@dkinzer I know it wasn't mentioned in this ticket, but I wonder if the same thing happens for web content.

dkinzer:
@sensei100 I checked.  It doesn't because web content doesn't have a show page on the app (it gets redirected to the website).

quoideneuf (approved):

PR #2213, opened 2020-10-21
BL-1166: Add hierarchical linking to subjects.

dkinzer:
For subject_display fields with subdivisions (i.e. those that we have
concatenated together using an em dash), add link options for searching
each level of the subject heading using the subdivisions on the full
record page

### DIFF
--- a/app/helpers/catalog_helper.rb
+++ b/app/helpers/catalog_helper.rb
@@ -371,10 +371,33 @@ module CatalogHelper
     creator
   end
 
+  # [a, b, c] => [[a], [a, b], [a, b, c]]
+  def hierarchies(array)
+    count = 0
+    array.reduce([]) { |acc, value| acc << array.slice(0, count += 1) }
+  end
+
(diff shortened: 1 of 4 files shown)


app/helpers/catalog_helper.rb
```diff
@@ -371,10 +371,33 @@ def creator_index_separator(args)
...
+  end
+
   def subject_links(args)
-    args[:document][args[:field]].uniq.map do |subject|
-      link_to(subject.sub("— — ", "— "), "#{base_path}?f[subject_facet][]=#{CGI.escape subject}")
-    end
+    separator = content_tag(:span, content_tag(:span, " — ", class: ""), class: "subject-level")
+
+    args[:document][args[:field]].uniq
+      .map { |subj| subj.sub("— — ", "— ") }
```
dkinzer, line 397: For another day, but, I'm curious whether or not we need this step anymore (assuming our data is pretty clean by now).

quoideneuf (approved):

PR #2225, opened 2020-10-26
Fix passing format=json throws error.

dkinzer:
Passing format=json as an argument to /everything? query will throw an error instead of loading results in JSON format.


dkinzer:
Decided to use this in another PR so closing.

PR #2231, opened 2020-10-27
BL-1355 Add banner

sensei100:
- Adds banner to LibrarySearch about the bookbot maintenance


dkinzer (changes requested):
@sensei100 This branch needs to be off of the latest release so that we can do the hotfix deploy without deploying anything else:

https://github.com/tulibraries/tul_cob/compare/v1.4.2..BL-1355-add-banner

PR #2234, opened 2020-10-29
[BL-1341] Create relevance tests for libguides.

dkinzer:
* Skip running on CI
* Idea is to trigger this on QA or locally but referencing prod Solr.


dkinzer:
FYI I have a script for generating the libguides relevance tests.  But it still need some TLC before it's ready for commiting.

bibliotechy (approved):

PR #2237, opened 2020-10-29
BL-1350 LC Display Field

sensei100:
- Adds lc_call_number field to search results when sorting by lc classification.
- Adds a box to the top of the search results that displays to the user that they are using lc sorting.
- Box includes a link to remove lc sort

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -646,8 +646,4 @@ class CatalogController < ApplicationController
     def set_thread_request
       LogUtils.request = request
     end
-
-    def lc_sort_selected?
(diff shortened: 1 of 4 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -641,4 +646,8 @@ def advanced_controller?
     def set_thread_request
       LogUtils.request = request
     end
+
+    def lc_sort_selected?
```
dkinzer, line 650: Looks like lc_sort_selected? is being defined twice.

sensei100: @dkinzer Thanks for catching that.  I removed the extra.

dkinzer (approved):

PR #2248, opened 2020-11-02
remove lc classification from adv search dropdown

quoideneuf:


app/views/advanced/_advanced_search_facets.html.erb
```diff
@@ -1,6 +1,6 @@
 <div class="container" id="advanced-search-facets" style="margin-top: 14px;">
 <% facets_from_request(facet_field_names, @response).each do |display_facet| %>
-  <% if should_render_facet?(display_facet) %>
+  <% if should_render_facet?(display_facet) && display_facet.name != 'lc_facet' %>
```
dkinzer, line 3: could this be handled via lc_facet's :if config?

quoideneuf: @dkinzer I'm not sure - just worried that should_render_facet? is used elsewhere. Anyway, I need to repost this because I used the wrong fork.

PR #2279, opened 2020-11-23
Bl 1366 remove podiatry pickups

quoideneuf:


sensei100 (commented):
@quoideneuf The changes that you made will only be reflected for requests from the bookbot.  If they also want regular hold requests to exclude PODIATRY, you will need to remove it from def self.possible_pickup_locations on line 25. I haven't read the ticket, so I'm not sure what the requirements are.

quoideneuf:
@sensei100 Ok, thanks. I removed it on that line as well.

sensei100 (approved):

PR #2290, opened 2020-12-01
BL-1372 don't exclude MAIN campus from valid pickup locations

quoideneuf:


sensei100 (commented):
@quoideneuf I just want to make sure this is the desired behavior.  Remote storage and Presser are actually different libraries than Charles.  Do they want those items from other libraries to be picked up at Main?  Does this resolve the issue with items on physical shelves within Charles?

quoideneuf:
@sensei100 I think I see what you meant. I changed it to use the library code rather than the campus code.

quoideneuf:
Will reopen shortly.

PR #2293, opened 2020-12-02
Authenticate dockerhub pulls.

dkinzer:
dockerhub will begin to throttle non anthenticated pulls so we need to
authenticate our pulls now.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,6 +42,9 @@ jobs:
   run_tests:
     docker:
       - image: cimg/base:2020.01
+        auth:
+          username: $DOCKERHUB_USER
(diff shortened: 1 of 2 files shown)


sensei100 (approved):

quoideneuf (approved):

sensei100:
@dkinzer Will there be a follow-up PR in the playbook for the env. variables?

dkinzer:
@sensei100 I added those directly to circleci via environment variable dashboard.

sensei100:
So we will just add that locally for the makefile command?

dkinzer:
@sensei100 so locally this wont be an issue because we wont hit the throttle limit ... the login will only happen if CI=true which it does in the circleci environment.

PR #2296, opened 2020-12-03
respond with not found for all random requests

quoideneuf:
See:
https://app.honeybadger.io/projects/56250/faults/71188759

What happened:
The app raised a routing error, and because it was in the production environment, tried to render an error page while recovering and then errored while rendering due to an environment variable that wasn't there being expected by the template. I have fixed both issues (I hope!).

image

### DIFF
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -11,6 +11,6 @@ class ErrorsController < ApplicationController
 
   private
     def layout(options)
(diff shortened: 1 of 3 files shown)


dkinzer (changes requested):
@quoideneuf looks good but could you please update this ticket with information relavent information from honeybadger or a reference an issue with that information.  Unfortunately honeybadger reports go away after a period of X weeks

dkinzer (approved):

PR #2304, opened 2020-12-07
BL-1373 conditionally render lc classification metadata

quoideneuf:
- abstracted logic for displaying lc classification to a controller concern
- created a local IndexPresenter class to customize fields_to_render so we only have to iterate through field presenter creation one time.


quoideneuf:
noticed something - will repost.

PR #2305, opened 2020-12-07
BL-1373 conditionally render lc classification metadata

quoideneuf:
Repost of: https://github.com/tulibraries/tul_cob/pull/2304

- abstracted logic for displaying lc classification to a controller concern
- created a local IndexPresenter class to customize fields_to_render so we only have to iterate through field presenter creation one time.


sensei100 (approved):

dkinzer:
oops I was in the middle of making a suggestion for this?

PR #2316, opened 2020-12-09
Bl 1364 add email address

quoideneuf:
Pulls the rapid request email addressee into a config file.
Also wraps a line in the mail constructor in a begin / rescue to protect against errors arising in the alma gem when the item has no holdings.
Unrelated change to prevent serving an html page for missing asset files.

### DIFF
--- a/app/models/concerns/blacklight/document/email.rb
+++ b/app/models/concerns/blacklight/document/email.rb
@@ -16,7 +16,11 @@ module Blacklight::Document::Email
         body << I18n.t(label, value: value.join("; ").gsub("|", "; "))
       end
     end
-    body << add_holdings_information
+    begin
+      body << add_holdings_information
+    rescue Alma::BibItemSet::ResponseError => exception
+      Honeybadger.notify(exception.message)
+    end
     return body.join("\n") unless body.empty?
   end
 
--- a/config/email_groups.yml
+++ b/config/email_groups.yml
@@ -1,0 +1,16 @@
+default: &default
+
+test:
+  <<: *default
+
+development:
+  <<: *default
+  RAPID_REQUEST:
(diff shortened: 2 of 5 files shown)


app/models/concerns/blacklight/document/email.rb
```diff
@@ -16,7 +16,9 @@ def to_email_text
         body << I18n.t(label, value: value.join("; ").gsub("|", "; "))
       end
     end
-    body << add_holdings_information
+    begin
+      body << add_holdings_information
```
dkinzer, line 20: @quoideneuf should we add a Honeybadger log for this rescue?

config/email_groups.yml
```diff
@@ -0,0 +1,16 @@
+default: &default
+
+test:
+  <<: *default
+
+development:
+  <<: *default
+  RAPID_REQUEST:
+    - librarystaff1@mailinator.com
+    - librarystaff2@mailinator.com
```
dkinzer, line 10: ooh, did not know about mailinator.com ... devshare opportunity? 😂

app/models/concerns/blacklight/document/email.rb
quoideneuf: @dkinzer I added it. I am not entirely sure what is going on here, but it will be interesting to see if notifies every time or never. As far as I can tell, items that can be rapid requested don't have holdings, and items without holdings will throw an error here. But I'm not totally sure. 
Example in HB https://app.honeybadger.io/projects/72785/faults/71604511

config/email_groups.yml
quoideneuf: 👍

dkinzer (approved):

PR #2318, opened 2020-12-09
BL-1298: Do not count article bookmarks.

dkinzer:
* Also removes more unnecessary article bookmark code.
* Fixes one styling issue that was causing font to be white in select button.

### DIFF
--- a/app/assets/stylesheets/partials/_bookmarks.scss
+++ b/app/assets/stylesheets/partials/_bookmarks.scss
@@ -25,10 +25,10 @@
   color: $red;
 }
 
-.clear-bookmarks {
-  font-size: 0.875rem;
+a.clear-bookmarks {
+  //font-size: 0.875rem;
   margin-left: 0.625rem;
-  margin-top: -0.225rem;
+  //margin-top: -0.225rem;
   padding: 0.4725rem 0.625rem 0.4125rem 0.625rem;
 
   @media(max-width: 601px) {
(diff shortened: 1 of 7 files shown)


sensei100 (changes requested):
@dkinzer This looks good, but there is one css issue on mobile devices with the send to menu. This is happening on production too.
Screen Shot 2020-12-10 at 9 17 07 AM

dkinzer:
@sensei100 since that issue is happening on production too, should we create a new ticket to resolve that one?

sensei100:
@dkinzer My opinion is that since this is an issue directly related to the work in this ticket, it makes sense to address it here.  Up to you.

dkinzer:
@sensei100 what device and browser has this issue plese.

sensei100:
@dkinzer I was using the chrome dev tools and clicking on the toggle device button

dkinzer:
@sensei100 I did that and I was not able to reproduce issue.  I was going to try browserstack next.

sensei100:
@dkinzer I see it on my iPhone 10 and I also see it when using safari's responsive mode.

Gemfile.lock
```diff
@@ -451,7 +451,7 @@ GEM
       rspec-mocks (~> 3.9)
       rspec-support (~> 3.9)
     rspec-support (3.10.0)
-    rubocop (1.5.2)
```
dkinzer, line 454: confused by this since we already have that upstream https://github.com/tulibraries/tul_cob/pull/2321/files

sensei100 (approved):

PR #2332, opened 2020-12-15
BL-1298 styling changes.

dkinzer:
* Update cite and sendto styling to match other buttons on bookmarks and show page.

### DIFF
--- a/app/assets/stylesheets/partials/_header.scss
+++ b/app/assets/stylesheets/partials/_header.scss
@@ -263,10 +263,10 @@ a#bookmarks_nav {
 
 #sendto-menu a.dropdown-toggle{
  	background-color: $white;
-  border: 0.0625rem solid $header-grey;
+  border: 0.0625rem solid $dark-grey;
   color: $red;
-	font-size: 0.875rem;
-	padding: 0.4725rem 0.625rem;
+	font-size: 1rem;
+	padding: 0.3725rem 0.75rem;
 
 	@media(max-width: 768px) {
 		background-color: $red;
@@ -334,9 +334,10 @@ a#bookmarks_nav {
 
 #citeLink {
 	background-color: $white;
-	border: 0.0625rem solid $header-grey;
+	border: 0.0625rem solid #6c757d;
 	color: $red;
-	padding: 0.4725rem 0.625rem;
+  font-size: 1rem;
+	padding: 0.375rem 0.75rem;
 
 	@media (max-width: 768px) {
 		background-color: $red;
--- a/app/assets/stylesheets/partials/_vars.scss
+++ b/app/assets/stylesheets/partials/_vars.scss
@@ -12,7 +12,7 @@ $facet-limit: #DDD;
 $grey: #333;
 $light-grey: #f4f6f7;
 $med-grey: #444;
-$dark-grey: #666;
+$dark-grey: #6c757d;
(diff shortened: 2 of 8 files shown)


app/assets/stylesheets/partials/_header.scss
```diff
@@ -263,10 +263,10 @@ a#bookmarks_nav {
 
 #sendto-menu a.dropdown-toggle{
  	background-color: $white;
-  border: 0.0625rem solid $header-grey;
+  border: 0.0625rem solid $dark-grey;
```
sensei100, line 266: @dkinzer Several of the indentations on this file don't look right.  Are they displaying correctly on your local?

app/assets/stylesheets/partials/_vars.scss
```diff
@@ -12,7 +12,7 @@ $facet-limit: #DDD;
 $grey: #333;
 $light-grey: #f4f6f7;
 $med-grey: #444;
-$dark-grey: #666;
```
sensei100, line 15: Did you check that this variable is not being used other places?  I don't want to inadvertently
change colors that we aren't supposed to.

config/locales/en.yml
```diff
@@ -111,7 +111,7 @@ en:
       sms: "Text"
       refworks: Refworks
       cite_html: "Cite %{current_range}"
-
+      sendto: "Send To %{current_range}"
```
sensei100, line 114: Will adding this here also display ${current_range} on the individual record pages?

app/assets/stylesheets/partials/_vars.scss
dkinzer: I thought I updated that variable... maybe I forgot to commit....

sensei100: You did update it. My concern is whether other elements are using that variable that will also be changed.

config/locales/en.yml
dkinzer: No, the individual record page uses it's own template and does not use this local 

image

app/assets/stylesheets/partials/_vars.scss
dkinzer: as far as I know this was being used for the grey around the buttons.

config/locales/en.yml
sensei100: I can see that it is also used as a filter-separator in the facets.  Not sure about anywhere else.  I'm just not sure we should change a variable when it is used in multiple places.

app/assets/stylesheets/partials/_vars.scss
sensei100: I can see that it is also used as a filter-separator in the facets.  Not sure about anywhere else.  I'm just not sure we should change a variable when it is used in multiple places unless we know that they want everything updated to the new color.

app/assets/stylesheets/partials/_header.scss
dkinzer: Yeah they were looking right on my local.

image
image

app/assets/stylesheets/partials/_vars.scss
dkinzer: supposedly facets use it... but I can't see much of a difference.

sensei100 (approved):

PR #2346, opened 2020-12-17
Translate location in text message

sensei100:


dkinzer (approved):

dkinzer:
If we could add a test to that it would be good.

sensei100:
@dkinzer There are already tests for this method. I updated the existing test with the new information required.

PR #2386, opened 2021-01-25
Pin devise to branch

sensei100:
- Devise has not yet merged the PR to work with omniauth 2.  The maintainer recommends using this branch in the meantime.


dkinzer:
We wont be able to test this until it get's to qa .. maybe even stage

dkinzer (approved):

dkinzer:
Seems like this is causing us a little trouble. I'll revert and add a ticket to look into it further later.

PR #2455, opened 2021-03-11
BL-1414 Purchase order bug

sensei100:
- Adds Japan faculty to the list of users that can use the purchase order feature


ebtoner:
@sensei100 @dkinzer Before you merge this PR, let me quickly doublecheck that there aren't additional user groups that we want to add (Japan student, Rome affiliates, etc.) I'm just seeing the list of the user groups here now, so I didn't realize it was so limited before.

dkinzer (approved):

sensei100:
@dkinzer @ebtoner This one should be ready to merge now.

PR #2465, opened 2021-03-15
BL-1419 Add guest link to login page

sensei100:
- Adds a Continue as guest link
- Includes prepopulated fields for title, date, and edition if applicable

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -7,6 +7,7 @@ class SessionsController < Devise::SessionsController
 
   # Overrides Devise::SessionsController#new.
   def new
+    @document = SolrDocument.find(params[:redirect_to].rpartition("doc-").last) rescue [ nil, SolrDocument.new({}) ]
     no_cache unless request.xhr?.nil?
     super
   end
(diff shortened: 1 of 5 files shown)


dkinzer (approved):

app/controllers/sessions_controller.rb
```diff
@@ -7,6 +7,7 @@ class SessionsController < Devise::SessionsController
 
   # Overrides Devise::SessionsController#new.
   def new
+    @document = SolrDocument.find(params[:redirect_to].rpartition("doc-").last) rescue [ nil, SolrDocument.new({}) ]
```
dkinzer, line 10: @sensei100, sorry I missed this; but what does this do?

sensei100: @dkinzer It breaks the redirect_url into pieces so that I can get the mms_id.

dkinzer: it's just that a session controller shouldn't know anything about SolrDocument.  This can result in being very confusing down the line.

sensei100: The requirements are to auto-populate form fields with title information.  There is no way to do that without document information.

PR #2466, opened 2021-03-15
BL-1393-text-updates

sensei100:
- Updates text on request rapid access modal

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -502,7 +502,7 @@ GEM
       simplecov_json_formatter (~> 0.1)
     simplecov-html (0.12.3)
     simplecov_json_formatter (0.1.2)
-    skylight (5.0.0)
+    skylight (5.0.1)
       activesupport (>= 5.2.0)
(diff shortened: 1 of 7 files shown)


app/controllers/sessions_controller.rb
```diff
@@ -7,7 +7,6 @@ class SessionsController < Devise::SessionsController
 
   # Overrides Devise::SessionsController#new.
   def new
-    @document = SolrDocument.find(params[:redirect_to].rpartition("doc-").last) rescue [ nil, SolrDocument.new({}) ]
```
dkinzer, line 10: @sensei100 since PR title says "text-updates" this change and the "if" statement below are jumping out at me.  Would you please add a comment to the PR or commit about what that seemingly non text update change is about? Thanks.

sensei100: Ugh, that shouldn't be part of this PR.  I will cherry pick and try again.

PR #2468, opened 2021-03-15
BL-1422 Add ga tracker for libkey button

sensei100:


sensei100:
@dkinzer I'm not sure why the indentation looks off.  It doesn't look like that on my local.

dkinzer (approved):

PR #2470, opened 2021-03-15
BL-1421: Always supress journals in book media context.

dkinzer:
* Fixes books media bento totals not matching books media search totals.

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -44,7 +44,7 @@ class CatalogController < ApplicationController
     # config.repository_class = Blacklight::Solr::Repository
     #
     ## Class for converting Blacklight's url parameters to into request parameters for the search index
-    # config.search_builder_class = ::SearchBuilder
+    config.search_builder_class = ::BooksSearchBuilder
     #
     ## Model that maps search index responses to the blacklight response model
     # config.response_model = Blacklight::Solr::Response
--- a/app/controllers/web_content_controller.rb
+++ b/app/controllers/web_content_controller.rb
@@ -12,6 +12,7 @@ class WebContentController < CatalogController
     config.connection_config[:url] = config.connection_config[:web_content_url]
     config.track_search_session = false
     config.index.title_field = "web_title_display"
+    config.search_builder_class = ::SearchBuilder
     # Do not inherit default solr configs from the catalog.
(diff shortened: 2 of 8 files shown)


app/controllers/web_content_controller.rb
```diff
@@ -12,6 +12,7 @@ class WebContentController < CatalogController
     config.connection_config[:url] = config.connection_config[:web_content_url]
     config.track_search_session = false
     config.index.title_field = "web_title_display"
+    config.search_builder_class = ::SearchBuilder
```
sensei100, line 15: @dkinzer Do the other controllers already have the search_builder_class?  Just wondering why it is only added to web_content.

dkinzer: @sensei100 I thought about it but decided not to until we add a search builder function that is specific just to we_content.  As it stands now web content is using the our default SearchBuilder custom class with not other changes.

dkinzer: That is to say the class would be an alias of `SearchBuilder`.

dkinzer: Something else too, since this was a journals specific issue I didn't want to make any changes to web_content that weren't absolutely necessary.  Just to not confuse what's going on.

sensei100: I'm just wondering why that was added to web_content, but not databases.

dkinzer: Oh, sorry just realized I didn't exactly answer your question
> Do the other controllers already have the search_builder_class?
Yes, they do.  I added here to pin it but notice that for the CatalogController (which is basically Books and Media) I updated the search_buiilder_class too.

app/controllers/catalog_controller.rb
```diff
@@ -44,7 +44,7 @@ def self.with_libkey?
     # config.repository_class = Blacklight::Solr::Repository
     #
     ## Class for converting Blacklight's url parameters to into request parameters for the search index
-    # config.search_builder_class = ::SearchBuilder
+    config.search_builder_class = ::BooksSearchBuilder
```
dkinzer, line 47: @sensei100 here's the other spot I update the search_builder_class

sensei100 (approved):

PR #2483, opened 2021-03-23
BL-1242 Emergency alerts from manifold

sensei100:
- Updates the emergency alert banner to pull JSON from the library website
- Banner will only display when the for_header field is false


dkinzer:
Was thinking that this is something we might want to add using javascript so that it happens the client side only and doesn't add more latency than the page already has.  What do you think @sensei100 ?

sensei100:
@dkinzer I don't think latency is a huge concern with json output this small.  I would prefer to keep it the way it is.

dkinzer (approved):

PR #2486, opened 2021-03-24
BL-1432 Remove libkey feature flag

sensei100:


dkinzer (approved):

dkinzer:
@sensei100 it might just be easier to set `WITH_LIBKEY=true`  as an environment variable in the .env file.  Then it will default to being on and if we ever need to remove it for some reason we can just toggle the key back to false.

sensei100:
@dkinzer I don't like the idea of leaving unused code around.  I think once a feature is proven it is ready, we should remove the flags.  I don't understand the failing spec though.

PR #2502, opened 2021-03-31
BL-1133: Fix drop-down selection accessibility issue.

dkinzer:
The bootstrap-select library is interfering with the accessibility use
case of drop-down select menus.  Removing it from our app fixes the
issue.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -16,7 +16,6 @@ gem "uglifier", ">= 1.3.0"
 gem "coffee-rails", "~> 5.0"
 gem "autoprefixer-rails"
 gem "execjs"
-gem "bootstrap-select-rails"
 gem "bootstrap", "~> 4.6"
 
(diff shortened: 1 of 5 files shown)


sensei100:
@dkinzer That library is what we use to auto-fill facets in the advanced search forms.  I don't think they would want to lose that.

dkinzer:
OK, maybe playing around with the order could also do the trick or updating the library if possible. And if not maybe limiting it just advanced pages.  I'll look into it on Monday.

dkinzer:
@sensei100 OK, locally I got it working by updating the version of bootstrap select.  Also I removed the bootstrap-rails-gem because it seems to just need the yarn setup to work (locally).

sensei100 (approved):

sensei100:
@dkinzer I'm so glad you figured this one out!

PR #2514, opened 2021-04-14
BL-1444 Gitlab pipeline setup

sensei100:
- Adds gitlab-ci file
- Removes duplicate env variable


sensei100:
@dkinzer I left the deploy step in for now.  We can remove it once we have the other way set up.

PR #2573, opened 2021-05-10
Fix Sprokets::AssetsNotPrecompiled error in local instance.

dkinzer:
Getting this error locally and I resolved it via this solution: https://stackoverflow.com/a/51898048/256854


sensei100 (approved):

dkinzer:
There was a better solution that just involved getting rid of the tmp/cache folder

PR #2697, opened 2021-07-16
BL-1519 Record page query lists

sensei100:
- Adds creator and call number query_lists to the record pages for catalog and journal documents

### DIFF
--- a/spec/features/query_list_spec.rb
+++ b/spec/features/query_list_spec.rb
@@ -11,14 +11,14 @@ RSpec.feature "Query List" do
     let (:item) { fixtures.fetch("simple_search") }
 
     scenario "displays query lists for catalog" do
-      visit "catalog/#{item['doc_id']}"
+      visit "catalog/#{item['doc_id']}?query_list=true"
       expect(page).to have_text("Call Number Asc")
     end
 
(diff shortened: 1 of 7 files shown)


spec/features/query_list_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+      expect(page).to have_text("Call Number Asc")
+    end
+
+    let (:item) { fixtures.fetch("url_856_0") }
+
+    scenario "displays query lists for journals" do
+      visit "catalog/#{item['doc_id']}"
+      expect(page).to have_text("Call Number Desc")
+    end
+  end
```
dkinzer, line 24: @sensei100 should we add a scenario where we visit record pages that we don't want query lists to appear, or is that silly since eventually we'll add lists to those places too?

dkinzer (approved):

spec/features/query_list_spec.rb
sensei100: @dkinzer I was having a hard time figuring out how to test that since those pages aren't returning xml.  I wasn't sure it was worth it since we plan to eventually have lists there.

dkinzer:
We need to add `&amp;per_page=5` to the matcher to make those tests pass.

PR #2699, opened 2021-07-16
BL-1514: Refactor additional_title_link

dkinzer:
*  Updates method to not render a link if only relation is present
*  Adds some tests to this helper method.

### DIFF
--- a/app/helpers/catalog_helper.rb
+++ b/app/helpers/catalog_helper.rb
@@ -370,14 +370,13 @@ module CatalogHelper
   end
 
   def additional_title_link(args)
-    title = args[:document][args[:field]]
+    args[:document][args[:field]].map do |title_data|
+      title_data = JSON.parse(title_data)
 
-    title.map do |title_data|
-      begin
-        title_data = JSON.parse(title_data)
-        linked_subfields = title_data["title"]
-        relation_to_work_prefix = title_data["relation"]
-      end
+      linked_subfields = title_data["title"]
+      relation_to_work_prefix = title_data["relation"]
+      next if linked_subfields.blank?
+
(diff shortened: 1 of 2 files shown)


app/helpers/catalog_helper.rb
```diff
@@ -370,24 +370,14 @@ def list_with_links(args)
...
-      end
-      link = fielded_search(linked_subfields, args[:field])
-
-      content_tag(:li, class: "list_items") do
-        if relation_to_work_prefix.present?
-          link.prepend("#{relation_to_work_prefix} ")
-        else
-          link
-        end
+    args[:document][args[:field]]
```
sensei100, line 373: @dkinzer Have they updated the requirements of this work?  The whole point of the previous work relating to this ticket is that they do not want the relation to be a part of the link.  I think the he only thing that should be considered a relation is 246 subfield i.

dkinzer: Could be they updated the requirements of this work. @ebtoner could we talk about this.  Just based on the expected links  provided in the issue, I assumed we wanted everything to be a link including relations which is what I did.

sensei100: @dkinzer I'm pretty sure the work for this needs to take place in cob_index.  That is where we are defining what should be a relation or title and that is where the misapplication is occurring.  The only things that should actually be labeled as relation are 246 subfield i.

dkinzer: Thanks, there's still  abug that happens when title is nil. Which will need to be fixed in this method.  I'll connect with @htomren about the details of the indexing.

sensei100 (approved):

PR #2701, opened 2021-07-19
BL-1424: Do not propagate URL params in bento links.

dkinzer:
Prevents params from being propagated via bento_links.


sensei100:
@dkinzer Does this mean we will no longer be propagating the query between searches?

dkinzer:
@sensei100 this should only affect the "view all" links in the bento boxes.

sensei100 (approved):

PR #2739, opened 2021-07-28
BL-1521 experimenting with tile styling

ebtoner:


sensei100 (commented):
@ebtoner The only potential issue I see is with the fixed height for the footer. The reason it was larger is because some of the call numbers are quite large and go down to a second line.  In those cases, the text is being cut off now.
Screen Shot 2021-07-28 at 10 43 32 AM

sensei100:
@ebtoner Another thing we might want to consider is adding a max-width.  Right now, if there is only one item in the list it is filling up the whole width of the page.

ebtoner:
@sensei100 Thanks! I'll take another crack at it.

sensei100 (approved):

PR #2763, opened 2021-08-06
BL-1543: Aug 2021 Release Final updates.

dkinzer:
* Remove donor_info_display field
* Remove feature flag from query links


sensei100:
@dkinzer I'm pretty sure specs need to be updated because we removed the query list feature flag.

dkinzer:
Yup.  Thanks.

sensei100 (commented):
The feature spec/query_list specs have this:       
visit "catalog/#{item['doc_id']}?query_list=true"
Should we remove those too?

dkinzer:
@sensei100 I think it's ok to leave those

sensei100 (approved):

PR #2775, opened 2021-08-12
BL-1544 Update ezborrow link

sensei100:


ebtoner:
@sensei100 for both changes, there is a space before the url value. Is that going to cause any issues?

sensei100:
@ebtoner I removed them just to be sure.

ebtoner (approved):

PR #2776, opened 2021-08-12
Devo 396 fix deprecation warnings part one

sensei100:
- Updates search bar to use component
- Uses document_presenter instead of index_presenter


dkinzer:
I checked this locally and the search query term disappears after you run the query.

dkinzer (approved):

PR #2905, opened 2021-11-01
BL-1562 Log purchase order requests

sensei100:
- Adds existing do_with_json_logger method to the purchase order requests 
- Fixes bug with the button where the text was not displaying when a user was logged in

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -548,10 +548,11 @@ class CatalogController < ApplicationController
     from = { email: email, name: name }
 
     mail = PurchaseOrderMailer.purchase_order(document, { from: from, message: params[:message] }, url_options)
+    log = { type: "purchase_order", user: current_user.id, mms_id: document.id }
     if mail.respond_to? :deliver_now
-      mail.deliver_now
+      do_with_json_logger(log) { mail.deliver_now }
     else
-      mail.deliver
+      do_with_json_logger(log) { mail.deliver }
     end
 
     redirect_back(fallback_location: root_path, success: "Your request has been submitted.")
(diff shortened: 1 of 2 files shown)


ebtoner:
@sensei100 Where does this text "Open a modal form to request a purchase for this item" appear? We no longer use the phase "purchase" in the UI to describe this service, so it seems outdated to me.

sensei100:
@ebtoner That text is not displayed anywhere.  This is to fix this issue that I am seeing on the request rapid access button
Screen Shot 2021-11-01 at 12 42 14 PM
:

ebtoner:
Ok, I understand now.  However, I'm not seeing that issue when logged in and can view that button text on prod without adding "text-white" to the div. 

image

I think I've encountered this issue before. I think it might occur to a user only after having submitted the paid access request. Did you submit a request for this item? If so, maybe only you see it?

sensei100:
@ebtoner Yes, I did submit a request. But wouldn't we want it fixed in that instance as well?

ebtoner:
We should definitely fix it. I'm wondering, though, if we should fix it via the Blacklight stylesheet class, since that seems to be the source of the problem, rather than by adding the bootstrap class. It looks like the problem is fixed if we switch from "btn-danger" to "btn-primary" (in  app/assets/stylesheets/partials/_colors.scss).  Perhaps sometime we can have a pairing session sometime to discuss when we should use a stylesheet vs. bootstrap?

sensei100:
@ebtoner I just prefer to use the built-in bootstrap classes when they are available.  I didn't think about changing the bin class since I assumed we had that specified for a reason.  Would you prefer that I take this fix out for now?

ebtoner (approved):

PR #2906, opened 2021-11-01
BL-1561 Remove article export for bookmarks

sensei100:


ebtoner:
@sensei100 Should we also remove the following in en.yml?: 
https://github.com/tulibraries/tul_cob/blob/7f46c4d9d9da91a07740a5027cc957af1c1cf1a1/config/locales/en.yml#L43-L57
I know it's not needed, since we've already removed the code that retrieves this text, but seems like it would be good to cleanup?

sensei100:
@ebtoner Sorry, I did that locally and didn't include it in the commit.  I added it now.

ebtoner (approved):

PR #2976, opened 2021-11-24
Fix stimulus-3 deprecation warnings

dkinzer:
Stimulus 3 has added a deprecation warning to data-target attributes because they are going away in future versions of stimulus.

### DIFF
--- a/app/views/catalog/_explanation_div.html.erb
+++ b/app/views/catalog/_explanation_div.html.erb
@@ -1,5 +1,5 @@
 <% if !cookies.fetch("#{controller_name}Once", false) %>
-<div class="row bg-background-grey rounded explanation-div mx-1" data-controller="explanation" data-exploration-target="<%= controller_name %>">
+<div class="row bg-background-grey rounded explanation-div mx-1" data-controller="explanation" data-explanation-target="<%= controller_name %>">
   <div class="explantion-wrapper col-sm-12 col-lg-10 mt-4 pl-5 pr-5 mx-auto" >
     <p class="explanation">
       <%= explanation_translations(controller_name) %>
--- a/app/views/catalog/_explanation_div.html.erb
+++ b/app/views/catalog/_explanation_div.html.erb
@@ -1,5 +1,5 @@
 <% if !cookies.fetch("#{controller_name}Once", false) %>
-<div class="row bg-background-grey rounded explanation-div mx-1" data-controller="explanation" data-target="<%= "explanation.#{controller_name}" %>">
+<div class="row bg-background-grey rounded explanation-div mx-1" data-controller="explanation" data-exploration-target="<%= controller_name %>">
   <div class="explantion-wrapper col-sm-12 col-lg-10 mt-4 pl-5 pr-5 mx-auto" >
     <p class="explanation">
       <%= explanation_translations(controller_name) %>
--- a/app/views/catalog/_index_availability_section.html.erb
+++ b/app/views/catalog/_index_availability_section.html.erb
@@ -7,12 +7,12 @@
     <div class="row button-break pt-md-3"></div>
     <div data-controller="availability" data-availability-url="<%= item_url(document.alma_availability_mms_ids.first, params.merge(doc_id: document.id, redirect_to: request.url).except(:page)) %>">
         <div class="controls">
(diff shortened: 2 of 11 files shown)


app/views/catalog/_explanation_div.html.erb
```diff
@@ -1,5 +1,5 @@
 <% if !cookies.fetch("#{controller_name}Once", false) %>
-<div class="row bg-background-grey rounded explanation-div mx-1" data-controller="explanation" data-target="<%= "explanation.#{controller_name}" %>">
+<div class="row bg-background-grey rounded explanation-div mx-1" data-controller="explanation" data-exploration-target="<%= controller_name %>">
```
sensei100, line 2:
```suggestion
<div class="row bg-background-grey rounded explanation-div mx-1" data-controller="explanation" data-explanation-target="<%= controller_name %>">
```

app/views/catalog/_index_availability_section.html.erb
```diff
@@ -7,12 +7,12 @@
...
+          <button data-action="availability#item" data-availability-ids="<%= document.alma_availability_mms_ids.join(',') %>" class="btn btn-sm btn-default availability-toggle-details" data-toggle="collapse" data-target="#physical-document-<%=  document.id %>" data-availability-target="button" id="available_button-<%=  document.id %>" aria-expanded="false" aria-controls="physical-document-<%= document.id %>">
             <i class="fa fa-spinner" aria-busy="true" aria-live="polite"></i>
             <span>Loading...</span>
           </button>
 
-          <div id="requests-container-<%= document.id %>" class="hidden requests-container float-right" data-controller="requests" data-target="show.request, availability.request">
+          <div id="requests-container-<%= document.id %>" class="hidden requests-container float-right" data-controller="requests" data-target="show.request" data-availability-target="request">
```
sensei100, line 15: @dkinzer I think the show.request target needs to be updated as well.

app/views/catalog/_lib_guide_recommender_catalog.html.erb
```diff
@@ -1,12 +1,12 @@
 <div class="lib-guides-recommender-catalog w-100 my-3 px-3">
   <div class="lib-guides-recommender-catalog-body rounded-top">
     <h2 class="lib-guides-recommender-catalog-heading text-red mb-0">Help &amp; Guides</h2>
-    <div class="d-flex justify-content-between pt-md-1 pb-md-4 px-md-4 catalog-guides" data-target="lib-guides.card">
-      <div class="lib-guides-recommender-librarians flex-fill p-4 rounded" data-target="lib-guides.flex">
+    <div class="d-flex justify-content-between pt-md-1 pb-md-4 px-md-4 catalog-guides" data-libg-guides-target="card">
```
sensei100, line 4:
```suggestion
    <div class="d-flex justify-content-between pt-md-1 pb-md-4 px-md-4 catalog-guides" data-lib-guides-target="card">
```

app/views/catalog/_lib_guide_recommender_catalog.html.erb
```diff
@@ -1,12 +1,12 @@
 <div class="lib-guides-recommender-catalog w-100 my-3 px-3">
   <div class="lib-guides-recommender-catalog-body rounded-top">
     <h2 class="lib-guides-recommender-catalog-heading text-red mb-0">Help &amp; Guides</h2>
-    <div class="d-flex justify-content-between pt-md-1 pb-md-4 px-md-4 catalog-guides" data-target="lib-guides.card">
-      <div class="lib-guides-recommender-librarians flex-fill p-4 rounded" data-target="lib-guides.flex">
```
sensei100, line 5: Is there a reason to remove the flex target?

sensei100 (changes requested):
There are a couple of suggestions, and I think one that was missed.

app/views/catalog/_index_availability_section.html.erb
dkinzer: @sensei100 I didn't see this get flagged in the javascript console.  I think we have some targets defined don't actually ever get used.  I wanted to leave them as is to see if we can further clean this up by removing unnecessary data tags.

Can you check to see if that's actually getting used.  I couldn't see where it was getting used.

sensei100, line 15:
```suggestion
          <div id="requests-container-<%= document.id %>" class="hidden requests-container float-right" data-target="show.request" data-availability-target="request">
```

sensei100: @dkinzer I'm not seeing the deprecation warning either, so I'm okay with leaving it for now to see if we can remove it.  I do notice that we no longer have a requests data-controller, so that should be removed.

sensei100 (approved):

PR #2989, opened 2021-12-07
BL-1570 Anchored link bug caused by nav-tools logic

ebtoner:
Anchored link bug with JS occurs only when "back to search" and "start over" buttons are present, but not previous-next navigation. I think the issue may be due to a mistake in my logic for https://github.com/tulibraries/tul_cob/pull/2936.


ebtoner:
@sensei100 Upon further inspection, I don't think this fixed the issue in BL-1570, but I think it's still beneficial to merge this fix for #2936.

sensei100 (approved):

PR #2996, opened 2021-12-08
BL-255 Fix query for creator link on full record page

ebtoner:


dkinzer (approved):
I don't think you actually need the utf8 param defined here to make this feature work.  But I guess it doesn't hurt either.

ebtoner:
@dkinzer That's probably true, but I was trying to replicate the search behavior found elsewhere when searching for creator_t

PR #2999, opened 2021-12-09
BL-1598: Add user#pay action.

dkinzer:


app/controllers/users_controller.rb
```diff
@@ -15,6 +17,38 @@ def account
...
   end
 
+  def pay
+    log = { type: "alma_pay", user: current_user.id, transActionStatus: params["transActionStatus"] }
+
+    type, message = do_with_json_logger(log) {
+      case params["transActionStatus"]
+      when "1"
+        alma_user = Alma::User.find(current_user.uid);
+        if alma_user.fines.send_payment
```
sensei100, line 27: @dkinzer This is only based on the user, not on fines.
```suggestion
        if alma_user.send_payment
```

PR #3005, opened 2021-12-10
BL-1600: report manifold pull alert errors to Honeybadger.

dkinzer:


app/controllers/application_controller.rb
```diff
@@ -49,7 +49,13 @@ def get_manifold_alerts
 
     @manifold_alerts_thread ||= Thread.new {
       Rails.cache.fetch("manifold_alerts", expires_in: 5.minutes) do
-        (HTTParty.get(alert_url, timeout: 1) rescue {})["data"] || []
+        resp = begin
+                 HTTParty.get(alert_url, timeout: 1)
+               rescue => e
+                 Honeybadger.notify(e)
```
sensei100, line 55: @dkinzer Is this just to make the manifold alert error explicit?  Because we have already been getting these errors in honey badger.

dkinzer: @sensei100 we have not been getting this particular error.  What we get is the error downstream after the fact . (i.e.) we've provided a default data of and empty array and something tries to use that and fails.

sensei100 (approved):

PR #3016, opened 2021-12-17
BL-1606: Add Online Payment feature.

dkinzer:
REF BL-1606


app/views/users/_user_fines.html.erb
```diff
@@ -26,7 +26,7 @@
       <% end %>
   </table>
   <div class="fines-link text-center">
-    <%= t("account.fines_not_electronic") %>
```
sensei100, line 29: @dkinzer Can we also remove this from the translations file so that we don't have unused text there?

sensei100 (changes requested):
@dkinzer Looks great, I just thought one minor deletion makes sense.

app/views/users/_user_fines.html.erb
dkinzer: Hmm... I've been actually thinking of updating the helper method to include this verbiage under certain conditions but didn't move forward on that util some discussion is resolved. Can we add a TODO instead, because it end up being removing it now, just to add it right back in another PR?

sensei100 (approved):

PR #3058, opened 2022-01-20
BL-1608 Payment feature styling

sensei100:
- Updates look/text of button
- Updates/styles flash messages for success and errors


spec/controllers/quik_pay_spec.rb
```diff
@@ -87,7 +87,7 @@
           allow(Alma::User).to receive(:send_payment) { balance }
           get :quik_pay_callback, params: params
           expect(response).to redirect_to users_account_path
-          expect(flash[:success]).to eq("Your balance has been paid.");
+          expect(flash[:info]).to include("Your fees have been paid.");
```
dkinzer, line 90: Just an FYI, that I used :success instead of :info because :info flashes weren't showing up.  But I'm assuming that you made some other change to fix that.  LGTM otherwise so approving.

dkinzer (approved):

spec/controllers/quik_pay_spec.rb
sensei100: @dkinzer I changed it to info because of the coloring they used in the mockup.  It showed up for me when working locally, but thanks for the heads up.

sensei100: I'll change it to notice just to be safe.

PR #3139, opened 2022-03-15
BL-1621 Google Analytics Tracking

sensei100:
- Updates to classes, ids, and categories to better track click events


app/assets/javascripts/listeners.js
```diff
@@ -11,43 +11,52 @@ $(document).ready(function(){
...
-      {class: "temp-help-btn", category: "search-results"},
-      {class: "google-preview", category: "record-page"},
-      {class: "hathitrust-link", category: "search-results"},
-      {class: "libkey-btn", category: "search-results"},
+      {id: "bento_journals_header", category: "bento-results"},
+      {id: "emailLink", category: "record-management"},
+      {id: "smsLink", category: "record-management"},
+      {id: "citationLink", category: "record-management"},
+      {id: "risLink", category: "record-management"},
+      {class: "toggle-bookmark", category: "rrecord-management"},
```
ebtoner, line 43: Update `category: "rrecord-management"` to `category: "record-management"`

app/assets/javascripts/listeners.js
```diff
@@ -11,43 +11,52 @@ $(document).ready(function(){
...
+      {class: "query-list-title-link", category: "query-list"},
+      {class: "query-list-online", category: "fulfillment"},
+      {class: "bento-online", category: "fulfillment"},
+      {class: "online-btn", category: "fulfillment"},
+      {class: "search-results-request-btn", category: "fulfillment"},
```
ebtoner, line 48: This is still using `class: "search-results-request-btn"` which needs to be updated.

ebtoner (changes requested):
Thanks @sensei100! Here are some things that I noticed reviewing this code
- Is there a reason we need to add the new class "request-btn" when we already have the existing class "request-button"? I'd like to reduce duplication if possible.
- The new "_header" ids for the bento sections look good, but we need to remove the old trackers, I think. For example, we now have a tracker for `bento_books_and_media_header`, so we can remove the tracker for `bento_books_and_media`. Let me know if I'm misunderstanding something here, though. 
- Can we add a tracker for the query list "View More" link as well? 
- See also comments in code

app/assets/javascripts/listeners.js
sensei100: I thought we were keeping this class for the actual search-results-request-btn?  I kept that class in the code for the search results, but let me know if I am supposed to take that out.

ebtoner: You're absolutely right. Thanks for clairfying!

ebtoner:
@sensei100 Everything is looking great! One more detail that I just noticed: In app/helpers/application_helper.rb, I think that the different instances of `link_to "View all #{total} online items", url, class: "bento-full-results"` need to be `link_to "View all #{total} online items", url, class: "bento-online-results"`. What do you think?

ebtoner:
All the trackers look good! That last change looks like it requires a minor style update, too, though
Screen Shot 2022-03-17 at 5 27 48 PM

ebtoner (approved):

PR #3145, opened 2022-03-18
BL-1591 Record page nav tools

sensei100:
- Moves the logic for this out of javascript
- Renders template based on size of the screen using css classes

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -42,17 +42,11 @@ $(window).on('turbolinks:load', function() {
 	}
 
 	if ($(window).width() < 768) {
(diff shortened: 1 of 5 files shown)


sensei100:
@ebtoner I tried adding a set height to the buttons.  I'm having difficulty reproducing this locally, so I'm hoping that it works.

ebtoner:
@sensei100 -- that addition actually ended up creating more issues with the display. Let's revert that last change and merge to QA. I can do more testing there and open a follow-up PR if needed. That said, it's very minor issue, so I may just leave as is. 

Screen Shot 2022-03-21 at 2 39 57 PM

ebtoner (approved):

PR #3159, opened 2022-03-30
BL-1617 Update layout to use library location for availability display

ebtoner:


app/helpers/availability_helper.rb
```diff
@@ -131,11 +131,11 @@ def location(item)
...
 
   def location_status(item)
-    location_name_from_short_code(item)
+    Rails.configuration.locations.dig(library(item), location(item)) || location(item)
   end
 
-  def location_name_from_short_code(item)
-    Rails.configuration.locations.dig(library(item), location(item)) || location(item)
+  def location_name_from_short_code(library_code, location_code)
+    Rails.configuration.locations.dig(library_code, location_code) || location_code
```
dkinzer, line 138: This looks great! But I wonder if this method should now be renamed since it requires both library_code and location_code to work.  Maybe just "get_location_name"? 

Also if most things are in the stack we could default that `location_code = "stack"`.

But I think overall this is a very good PR and I'm happy to merge it as is.

dkinzer (approved):

app/helpers/availability_helper.rb
ebtoner: Thanks David! I really appreciate your feedback! 

I'll switch it from "location_name_from_short_code" to "location_name_from_short_codes" to indicate that multiple codes are needed. I don't want to deviate too much from the current name.

This method already defaults to displaying the raw location code. If there is no location code, I don't think that we should default to "stacks" as it might provide misleading information to our users. It's better that it come back as nil in that scenario.

PR #3242, opened 2022-06-08
Allow local development using m1 mac.

dkinzer:
This adds m1 platform specific changes to allow local deevelopment using
m1 computer.  This does not change how application works for non m1 macs
etc.

This fixes m1 issues for building production image using
`make build` and `make run`

This also fixes issue for regular local dev using `bundle install`

This does not fix issue for docker-compose in local development.  That
is still a TODO.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,16 +67,17 @@ ci-bundle-install:
 ci-yarn-install:
 	$(DOCKER) exec app yarn install --frozen-lockfile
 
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -1,7 +1,7 @@
 #Defaults
 include .env
 export #exports the .env variables
-
+h
```
sensei100, line 4: @dkinzer Is this just a typo?

dkinzer: oops... thanks!

PR #3280, opened 2022-07-12
Don't use the listen gem in development

sensei100:


dkinzer (approved):

config/environments/development.rb
```diff
@@ -85,7 +85,7 @@
 
   # Use an evented file watcher to asynchronously detect changes in source code,
   # routes, locales, etc. This feature depends on the listen gem.
-  config.file_watcher = ActiveSupport::EventedFileUpdateChecker
```
dkinzer, line 88: I never realized we are automatically watching files in development. I think just this line would have fixed the issue since watch doesn't usually get invoked automatically.  But then again as far as I know non of us uses this feature.

PR #3329, opened 2022-08-22
BL-1642 Update python version in CircleCI config

sensei100:


dkinzer (approved):

dkinzer:
I think this is dependent on an update to the tul_cob_playbook version: https://github.com/tulibraries/tul_cob_playbook/blob/qa/.python-version


But approving now assuming that will happen later.

PR #3330, opened 2022-08-22
BL-1663: Disable the sms feature.

dkinzer:
Since we can't guarantee that the sms feature will work consistantly across providers, then it's better to disable the feature than to provide a half broken one.


dkinzer:
Closing in favor of phasing out with warning.

PR #3337, opened 2022-08-25
BL-1665 Update colors to better match website

sensei100:
Updates colors to match the new style guide.
Removes colors that we are no longer using.
In cases where the color we had defined was different than the library website, the color was updated to match the website.


config/initializers/assets.rb
```diff
@@ -21,4 +21,4 @@
 # Precompile additional assets.
 # application.js, application.css, and all non-JS/CSS in the app/assets
 # folder are already added.
-# Rails.application.config.assets.precompile += %w( admin.js admin.css )
+Rails.application.config.assets.precompile += %w( faustina.woff2 roboto-condensed.woff2)
```
ebtoner, line 24: @sensei100 Did you intend to include this update in this PR? I know that the local precompiling of assets can be a bit finicky at times, so I suspect that this is related to that. Before merging, I just want to doublecheck that this should be included in the main app.

sensei100: @ebtoner I think that was a result of my last PR.  Locally, I get an error related to preloading the font files.  Adding them to this file allowed me to move forward.

ebtoner (approved):

PR #3365, opened 2022-08-30
Update blacklight-marc to 8.0.0

sensei100:
It looks like this module was renamed about six months ago.  Both the old and new names work for me locally, but I am hoping updating it will fix the issue on QA that happened yesterday.  The error showed:

uninitialized constant Blacklight::Solr::Document::Marc
Did you mean? Blacklight::Marc

### DIFF
--- a/app/models/solr_document.rb
+++ b/app/models/solr_document.rb
@@ -35,7 +35,7 @@ class SolrDocument
   SolrDocument.use_extension(Blacklight::Document::Sms)
 
   # Add Blacklight-Marc extension:
-  SolrDocument.use_extension(Blacklight::Solr::Document::Marc) do |doc|
+  SolrDocument.use_extension(Blacklight::Marc::DocumentExtension) do |doc|
     doc.has_field? "marc_display_raw"
   end
   SolrDocument.extension_parameters[:marc_source_field] = "marc_display_raw"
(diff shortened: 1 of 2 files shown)


dkinzer:
I don't think this is correct.  In the blacklight-marc documentation they still say that to use you need to extend using `Blacklight::Solr::Document::Marc` and they have a spec that is working: https://github.com/projectblacklight/blacklight-marc/blob/main/spec/controllers/catalog_controller_spec.rb#L5-L7

app/models/solr_document.rb
```diff
@@ -35,7 +35,7 @@ class SolrDocument
   SolrDocument.use_extension(Blacklight::Document::Sms)
 
   # Add Blacklight-Marc extension:
-  SolrDocument.use_extension(Blacklight::Solr::Document::Marc) do |doc|
```
dkinzer, line 38: I don't think this is correct. In the blacklight-marc documentation they still say that to use you need to extend using Blacklight::Solr::Document::Marc and they have a spec that is working: https://github.com/projectblacklight/blacklight-marc/blob/main/spec/controllers/catalog_controller_spec.rb#L5-L7

sensei100: @dkinzer that worked for me locally, but that is what is throwing an error on QA.  The actual model file has an example using this configuration: https://github.com/projectblacklight/blacklight-marc/blob/main/app/models/concerns/blacklight/marc/document_extension.rb#L24

I don't think they updated the readME.

dkinzer: @sensei100 Yup. That makes sense

dkinzer (approved):

PR #3377, opened 2022-08-31
BL-1673 remove extra whitespace on right in mobile view

cdoyle-temple:


sensei100 (commented):
@cdoyle-temple I'm not sure why all of this work is included in this ticket.  I don't think the footer files were part of this issue.

ebtoner:
@cdoyle-temple @sensei100 It looks like some of that may just be css formatting changes? Is that accurate?

sensei100:
> @cdoyle-temple @sensei100 It looks like some of that may just be css formatting changes? Is that accurate?

There are changes to the way that I had done the columns.
<main id="main-container" class="col-sm-10 col-md-12 mx-auto p-0 m-0" role="main">
<main id="main-container" class="col-12 col-md-10 m-0" role="main">

I have updated my editor to use 2 spaces instead of 4 for indentation, so that should be resolved going forward.

sensei100:
@ebtoner @cdoyle-temple The reason that I question this is because currently on Librarysearch, everything in the footer is in alignment in the mobile view, which is not the case currently in manifold. In the past, that has been something that has been requested.  @ebtoner maybe you can weigh in on whether that is desired behavior.

ebtoner:
@cdoyle-temple @sensei100 I agree that I like to keep our current column structure and mobile flexibility if we can. 

Also, I pulled down this code to take a quick look, and unfortunately, it doesn't solve the core issue. You can still scroll to the right on the mobile view. It's a red background instead of whitespace, but I was hoping to find a solution that prevented that type of overflow entirely 

Here's the main view: 
Screen Shot 2022-09-02 at 10 10 23 AM

Here's the view when you scroll to the right: 
Screen Shot 2022-09-02 at 10 10 28 AM

A hacky solution would be to add html, body { overflow-x: hidden; } but I'm not sure if that's a best practice.

ebtoner:
Actually, I think I figured out what the cause of the whitespace is! 

I has to do with margins for the .documentHeader, with Bootscrap row class. 
Screen Shot 2022-09-02 at 10 24 10 AM

@cdoyle-temple @sensei100 does that seem right?

sensei100:
@ebtoner That looks like it could be it.  Just so you are both aware, I have some code commented out at the bottom of the tuck.scss.erb file.  When you uncomment it out, it add some background/text colors that make it easier to see issues like this, especially in the mobile view.

cdoyle-temple:
I can look at if the footer changes are specifically needed, but all these changes combined is what it took for me to get no left/right scrolling in mobile view. @ebtoner I am not seeing any horizontal scrolling in any of the browsers. @sensei100 I'm not seeing where the flexibility of the footer has changed behavior.

cdoyle-temple:
@ebtoner Fixing the documentHeader to stay within the container was not enough to prevent the horizontal scrolling.

sensei100:
@cdoyle-temple I'm not sure why changes had to be made to shared/fooer.html.erb as part of this work.  I am talking about the alignment of footer content on small screens.  The first screenshot is library search and the second is manifold. In Librarysearch the donor button and social media buttons are in alignment with the Quick links content.

Screen Shot 2022-09-02 at 12 09 24 PM
218-03a5fff9-b660-4d42-8521-d9aedcd0bf1c.png)

cdoyle-temple:
@sensei100 Sorry for my confusion, it appears that the changes to the footer were not made as part of this PR. The only change to the footer here is an indentation correction. I'm looking into the cause of this behavior, and it appears to be that the container class around the links is causing a gradual shifting to the left as the browser window size decreases, as they should, but the footer header does not have a container class and so is already all the way to the left. If you resize to 576 pixels (bootstrap small breakpoint), you'll see them perfectly aligned. It's the intermediate sizes I'll look to correct.

cdoyle-temple:
@sensei100 Having that container only around the links section has the effect of the donate and social buttons being blown out past the container width on larger screens. I'll adjust so both footer sections stay in sync and add to this PR, unless you want a separate PR?
Screen Shot 2022-09-02 at 1 28 04 PM

sensei100:
@cdoyle-temple I am not able to recreate that behavior.  I think all of these extra changes are making it difficult to see what the actual changes for this ticket are.  I wish that you would have just told me that my spacing was off so that I could have fixed the files that I updated.

cdoyle-temple:
@sensei I haven't updated anything yet. The issue remains with your files I suppose. I will not continue to make a fix for it. FYI, using the inspector tool in chrome, you can set the view to mobile/responsive and then set any width you'd like. Higher values than your actual screen real estate will shrink the page content (zoom level) so it fits in the view.

app/assets/stylesheets/partials/_footer.scss
```diff
@@ -2,127 +2,129 @@
...
-	  transform: skewY(-2.5deg);
-	  @media screen and (min-width: 1599px) {
-		top: -55px;
-		height: 121px;
-	  }
-	}
-  }
-  .diagonal.footer-diagonal {
-	margin-top: 10rem;
+  background-color: $lighter-grey;
```
sensei100, line 113: @cdoyle-temple It seems like there is no indentation for this section now.

ebtoner (approved):

PR #3383, opened 2022-09-01
BL-1688: Use creator_ms in creator browser link.

dkinzer:


ebtoner:
@dkinzer you added reviewers to this PR, but it's still in draft mode. Is this ready for review?

dkinzer:
@ebtoner I'm not sure any of these changes we've been making so far are ultimately necessary to fix the underlying issue.  I am waiting until we can discuss the issue to determine if we even want to pursue this change.

PR #3387, opened 2022-09-02
BL-1674: Add friendly error message to articles timeout.

dkinzer:


ebtoner:
@dkinzer I'm thinking over the text for the error message. That may need to wait until Monday. Would you prefer that I merge this ticket now or wait until I have those revisions?

dkinzer:
I'm okay waiting until it's ready.

sensei100 (approved):

sensei100:
@ebtoner @dkinzer I approved this, but didn't merge since I wasn't sure whether the error message was ready.

ebtoner:
Here's the updated text:

- general error flash: "Your search query has timed out."
- additional pagination error flash: "You may have exceeded the maximum number of pages allowed for an Articles search query in Library Search."

We also already have https://github.com/tulibraries/tul_cob/blob/920df66499d1b1596456ae00040691fcd61d9031/config/locales/en.yml#L98. Can you update that to "You have exceeded the maximum number of pages allowed for a search query in Library Search."? (I can make that a separate PR, if preferred)

Also, should new error text also be in the en.yml file?

dkinzer:
@ebtoner OK. I'm not sure about the en.yml question though.  I think we can all have a discussion about this since technically those entries are there to make it easier to port our website to different languages but as far as I know we don't do this or plan to ever do it?

ebtoner:
We can set aside the en.yml for the purposes of this PR. I've been looking at the error messages more generally for another ticket and agree that there are some broader questions to discuss.  Let's just make the changes that you outlined with updated text for now.

PR #3404, opened 2022-09-07
Update footer-header for xl and md screen sizes

sensei100:


ebtoner (approved):

cdoyle-temple (commented):
@sensei100 Are the icons in the footer-header supposed to be indented like this at desktop sizes?
Screen Shot 2022-09-07 at 10 39 51 AM

ebtoner:
@cdoyle-temple I would prefer for there not to be that type of indentation. Instead, each should be aligned with each end of the container. 

@sensei100 Can you take a look at this?

PR #3489, opened 2022-11-04
Remove an unsued name paramter value in _index_header.html.erb

### LINKED ISSUE #3488: Why is `tag: :div` passed to render_index_doc_actions?
I was looking at this code https://github.com/tulibraries/tul_cob/blob/3469c4aae69ae7a383c8759be19d9c78e79f7b09/app/views/catalog/_index_header.html.erb#L20 and I couldn't figure out why the `tag:` argument is passed. Do you know what that is used for?

dkinzer:
Closes #3488

### DIFF
--- a/app/views/catalog/_index_header.html.erb
+++ b/app/views/catalog/_index_header.html.erb
@@ -17,7 +17,7 @@
               <h3 class="index_title">
                 <%= link_to_document document, counter: counter %>
               </h3>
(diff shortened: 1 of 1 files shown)


cdoyle-temple (approved):

dkinzer:
Thanks @cdoyle-temple I'm putting it back into draft because there's still an open question about whether or not it's actually getting used. I'm going to open up a debug session to verify it.

PR #3491, opened 2022-11-04
changed libc6 version

ChinDiesel:


dkinzer:
@ChinDiesel I built this locally and it's not going to be enough to fix the issue. Because there are multiple things that fail. Once we fix one the next one will fail and then the next one.  I opened up a PR that fixes all the then so we don't have to a bunch of PRs to get the thing running again: https://github.com/tulibraries/tul_cob/pull/3490

ChinDiesel:
@dkinzer yeah so it's kind of weird. I saw you assigned it to me so I thought you wanted me to check the versions. I was able to complete the build on my machine with just that one fix for lib6c but I saw in yours you had multiple. curious why I didn't see those

dkinzer:
You know what. I think I must have built on top of an old branch. If you got it to work with this one on your local. Thanks. I'm merging this and closing my other one.

dkinzer (approved):

PR #3508, opened 2022-11-12
BL-1729: Collapse multiple request options.

dkinzer:
Fixes bug with requesting bookbot media item for faculty and library
staff user groups.


ebtoner:
@dkinzer -- see my comment on the ticket BL-1729. I don't think this approach will address the core issue.

ebtoner:
@dkinzer - does your commit update address scenarios where there is more than 2 items?

dkinzer:
@ebtoner I believe so.

ebtoner (approved):

PR #3510, opened 2022-11-14
Rename check_for_full_http_link helper method.

dkinzer:
I was reading the code and was having a hard time understanding what
check_for_full_http_link method did based on name.  I think
build_availability makes more sense or maybe we can come up with another
more suitable name that doesn't imply that
check_for_full_http_link is checking that somethig is true or false
and therefor not actually rendering html that will be displayed on
a page.

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -182,7 +182,7 @@ class CatalogController < ApplicationController
     config.add_index_field "creator_display", label: "Author/Creator", helper_method: :creator_index_separator
     config.add_index_field "format", label: "Resource Type", raw: true, helper_method: :separate_formats
     config.add_index_field "lc_call_number_display", if: :render_lc_call_number_on_index?
-    config.add_index_field "url_finding_aid_display", label: "Finding Aid", helper_method: :check_for_full_http_link
+    config.add_index_field "url_finding_aid_display", label: "Finding Aid", helper_method: :build_availability
     config.add_index_field "availability"
     config.add_index_field "purchase_order_availability", field: "purchase_order", if: false, helper_method: :render_purchase_order_availability, with_po_link: true
     config.add_index_field "bound_with_ids", if: false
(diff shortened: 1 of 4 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -274,8 +274,8 @@ def self.libwizard_tutorial?
     config.add_show_field "lc_call_number_display", label: "LC Classification"
     config.add_show_field "alma_mms_display", label: "Catalog Record ID"
     config.add_show_field "language_display", label: "Language"
-    config.add_show_field "url_more_links_display", label: "Other Links", helper_method: :check_for_full_http_link
-    config.add_show_field "electronic_resource_display", label: "Availability", helper_method: :check_for_full_http_link, if: false
+    config.add_show_field "url_more_links_display", label: "Other Links", helper_method: :build_availability
+    config.add_show_field "electronic_resource_display", label: "Availability", helper_method: :build_availability, if: false
```
dkinzer, line 278: This doesn't even get rendered so I'm not even sure why are passing a helper_method to it.  But I guess that's a different issue than the naming.

ebtoner: @dkinzer I find build_availability to be more confusing, actually. This method is specific to electronic resource links. "Availability" has a much broader meaning in tul_cob beyond e-resources.

dkinzer: Yeah, I'm not at all happy with build_availability either. But I get confused by the current name too.

sensei100: It's been awhile since I looked at this code, but here is what I think is going on.  This field is not rendered on the show page, but the field is used to determine wether the electronic resource is a direct link, or a list of multiple resources.  The helper method is used to check whether the resource starts with "http".  If it does, then the direct link is rendered as opposed to the list of all of the electronic resources available.

dkinzer:
I'm closing this.  Maybe the next time I'm working in the same part of the code I may take another stab at it.

PR #3512, opened 2022-11-14
Remove confusing helper_method configuration.

dkinzer:
We're defining helper_method for some of the fields that we also set
`if:
false` to.

The :if config for blacklight is a configuration tha is used to set if
a field gets rendered by blacklight.  If so, then the defined
:helper_method would be used to render the field instead of whatever the
default is.  But since we are setting :if that means we are not
rendering this field (at least we are not having Blacklight render it
for us).

It's unnecessary to define a rendering :hlper_method field level config
if the field is configured never to be rendered by blacklight.


dkinzer:
There's something going on here that doesn't make sense to me but I don't have time to look into it.

PR #3543, opened 2022-11-21
BL-1699: Add axe rspec gem.

dkinzer:
* Also adds some tests but skips them because they fail.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -28,6 +28,7 @@ gem "jbuilder", "~> 2.11"
 
 
 group :development do
+  gem "axe-core-rspec"
   gem "flamegraph"
   gem "ruby-prof"
   gem "spring"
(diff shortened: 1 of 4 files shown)


ebtoner:
@dkinzer @sensei100 Should we wait to merge this until after the prod release? I'm wondering what testing / work the addition of this gem will require...

dkinzer:
@ebtoner I don't think it matters since it's a development gem that is only being invoked in the circleci context or locally if you run rpsec tests.

ebtoner:
@dkinzer I guess my concern is that it will flag a bunch of errors that we previously were unaware of. I'd like to at least finish out the other outstanding work for this sprint first before merging.

ebtoner:
Ok, that makes sense. Thanks for clarifying!

PR #3690, opened 2023-03-15
BL-1758 Update author/creator displays

sensei100:
- Adds back responsibility field values
- Adds comma separators to primo author index displays with multiple authors


ebtoner:
@sensei100 It looks like the view _show_fields.html.erb presents an empty <dt> for the repsonsibility fields, while the view _fields.html.erb remove the <dt> markup completely. I think we want the logic in _show_fields.html.erb for both?

sensei100:
@ebtoner I made the change you requested, but I don't see any difference in the view.
Screen Shot 2023-03-16 at 10 20 42 AM

ebtoner:
There shouldn't be any visual change, but it might make a difference to tools like screen readers.

PR #3708, opened 2023-03-27
BL-1741 Add pickup location configuration for Rome Campus Library

ebtoner:


sensei100:
@ebtoner Would we also need to add similar logic for Japan?

ebtoner:
@sensei100 -- that's a good point! I'm honestly not sure. Japan items are not requestable to most user groups, but for the ones which are (ex. Faculty/Admin - Japan), the same problem as the Rome items exists. I'll need to follow up with Cynthia about this, who is out until Wednesday.

sensei100:
@ebtoner Since she is out, would you prefer I merge this one now and we follow up with the Japan location in a separate PR?

PR #3728, opened 2023-03-30
Updates for new GA4 analytics

sensei100:


app/views/layouts/blacklight.html.erb
```diff
@@ -36,6 +36,19 @@
     <%= csrf_meta_tags %>
     <%= content_for(:head) %>
   </head>
+
+  <% if @nonce.present? %>
```
dkinzer, line 40: Are we using the @nonce generated value somewhere.  If not, why not just use if `Rails.env.production?` on it's own.

sensei100: @dkinzer Chris said that was necessary for the security scan when we eventually  go to production in kubernetes. He said it has to do with the content security policy.

dkinzer: It doesn't make sense to me but I trust you all know what you are doing.

dkinzer (approved):

PR #3738, opened 2023-03-31
BL-1758 Author/responsibility updates

sensei100:
* Remove responsibility_display from secondary fields on full record pages
* Updates multiple authors in articles to display on separate lines (this matches what we do for catalog results)

### DIFF
--- a/app/assets/stylesheets/partials/_index_results.scss
+++ b/app/assets/stylesheets/partials/_index_results.scss
@@ -236,3 +236,12 @@ dd.blacklight-format {
     border-top-left-radius: 0 !important;
   }
 }
+
+/* === PRIMO ARTICLE RESULTS === */
+dd.blacklight-creator {
+  @media(min-width: 768px) {
+    a:not(:first-child) {
+      margin-left: 120px;
+    }
+  }
+}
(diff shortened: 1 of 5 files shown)


ebtoner:
@sensei100 -- I need to think about multiple authors issue a bit more. Ideally, we'd have them joined by comma on the search results page and separated by a line break on the full record page. However, I understand that would add additional complexity here, so I'm going to review the UI/UX again.

sensei100:
@ebtoner I'm finding the lack of consistency confusing.  If we want multiple authors to display on separate lines in the catalog results, then it makes sense that we would do the same here.

ebtoner:
Yes, we want them to display consistently on the full record display between the catalog and articles. But we also want to maintain the displays we have on the search results page. 

If you look at the production site, this is the display for the Article search results: 
image

While this is the display for the Articles full record page: 
image

I'm guessing that adding the links to the author display in the Article search results makes this harder to achieve.

ebtoner:
@sensei100 -- Ideally, I would still like to maintain the displays from screenshots in my last comment. 

Could we do something like Stanford does in their code? See https://github.com/sul-dlss/SearchWorks/blob/60379f305e1cbfbf2cf98ff4a45f36c533be2238/app/controllers/articles_controller.rb#L111

If you don't have time to look into this, I'm happy to investigate further this week.

sensei100:
@ebtoner I would rather you picked this up.

ebtoner:
@sensei100 -- I made an update based on Stanford's approach. Please let me know, though, if you'd advise against this solution. I'm still learning the best practices for rails!

ebtoner (approved):

PR #3771, opened 2023-04-21
Increase mailer timeout.

dkinzer:
Allow up to one minute for mails to deliver via smtp.

This is an attempt to fix an increase in mail timeout errors.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -83,7 +83,8 @@ Rails.application.configure do
     user_name: "templelibraries@gmail.com",
     password: ENV["ASKTULIB_PASSWORD"],
     authentication: :plain,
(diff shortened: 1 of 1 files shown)


ebtoner:
@dkinzer According to https://edgeguides.rubyonrails.org/action_mailer_basics.html#action-mailer-configuration, the config labels for timeout settings are :open_timeout and :read_timeout. Am I missing something?

dkinzer:
Good point, my bad.

ebtoner:
Do we need to add :open_timeout too?

dkinzer:
I didn't see an error for the open connection timeout so I decided to take a wait and see approach to updating that value

ebtoner (approved):

PR #3961, opened 2023-09-15
DEVO-849: Enable SAST in gitlab pipeline.

dkinzer:


ChinDiesel (approved):

dkinzer:
@sensei100 unfortunately it is related.  There's a security issue with the function that I am getting rid of.  I think it's easier to just remove the code then fix the issue since we don't actually use shib anymore.  I think I finally got rid of all the shib code now with this last update 🤞🏽

PR #3995, opened 2023-10-09
DEVO-809-update-orbs-for-circle-ci

cdoyle-temple:


sensei100:
@cdoyle-temple I don't think that the update to the solr submodule should be part of this.  Can you take that file change out of the PR?

sensei100 (approved):

PR #4041, opened 2023-11-06
BL-1816 Remove unused online_only_checkbox partial

nomadicoder:
- This empty file nor it's former partial does not appear to be referenced in `tul_cob`.
- `app/views/primo_central/_online_only_checkbox.html.erb`  is an empty file.  Originally, it contained `render_online_only_checkbox` which is not referenced within `tul_cob`. I recommend removing the empty file.


nomadicoder:
Work in progress -- Checking to see if removal passes checks run within CircleCI

sensei100 (approved):

PR #4045, opened 2023-11-07
BL-1654 Add date field to query item displays

ebtoner:


sensei100:
@ebtoner There is a trick that we use for nil sometimes instead of the conditional.  You could put this and it would function the same way: 
<% doc["imprint_date_display"]&.each do |date| %>
The & is called the safe navigation operator and it lets you call methods on objects without worrying about the object being nil.

ebtoner:
Right! I knew about that operator, but wasn't thinking of it when I did this. Thanks for the suggestion! Other than that, just keep it all in the view partial?

sensei100:
@ebtoner Yes, I think it is fine to keep it here.

sensei100 (approved):

PR #4051, opened 2023-11-09
BL-1815 Reorganize helper methods

sensei100:
This PR adds new files for:
* Purchase order helpers
* Hathitrust helpers
* Electronic resource helpers
* Notes helpers

It also moves some methods related to availability from the catalog helper to the availability helper.


dkinzer (approved):

ebtoner:
@sensei100 All of the methods in note_helper.rb are related to electronic resource notes. If I was looking for those methods, I'd probably look at electronic_resource_helper.rb first. I see that David already approved this PR, which is fine, but I thought that I'd astill sk if it would make more sense to group those methods with the electronic resource methods? If not, could we rename note_helper.rb to be a bit more specific?

sensei100:
@ebtoner I can move them into the electronic resources helper file if that makes more sense.  I thought maybe the notes work was different enough that it belonged on its own.  Most of the methods in the electronic resource display are related to the links to the resources, but we can keep them coupled together.

ebtoner:
@sensei100 I do think that it makes sense to couple them together. The main method render_electronic_notes is only ever used in electronic resource helper methods.

PR #4056, opened 2023-11-10
BL-1808 Diagnose invalid session-id in CircleCI

nomadicoder:
Shrink resource size to create failing test


nomadicoder:
No longer needed

PR #4116, opened 2023-12-08
BL-1827 Update uniform title spec

ebtoner:
This needs to be updated because of changes in https://github.com/tulibraries/cob_index/pull/597


sensei100 (approved):

sensei100:
@ebtoner Has tul_cob been updated to use the cob_index change yet?  This test should pass if you update the gem.

PR #4227, opened 2024-02-16
BL-1823 Extend timeout time to 10 seconds

nomadicoder:
Addresses intermittent timeout with Primo Central

### DIFF
--- a/app/models/blacklight/primo_central/document.rb
+++ b/app/models/blacklight/primo_central/document.rb
@@ -156,7 +156,7 @@ module Blacklight::PrimoCentral::Document
 
       Thread.new {
         begin
-          HTTParty.get(libkey_articles_url, timeout: 2)["data"]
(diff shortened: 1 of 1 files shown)


dkinzer:
Just FYI, I think we specifically made this 2 seconds in order to prevent articles from causing heavy delays if the articles api is down or too slow.

ebtoner:
@nomadicoder -- In the description, you mention that this is for a Primo timeout. While this file is for the Primo article document, the actual code for the lookup in question is related to LibKey, which is a different third-party service that we use from Third Iron for the "View Full Text PDF" link. 

I think that David mentioned that increasing this timeout could potentially affect performance, so I hestiate to make the timeout 10 seconds. @dkinzer -- do you think a smaller increase of 1 or 2 seconds would be okay? Alternatively, we could just have this fail more gracefully.

ebtoner (approved):

PR #4242, opened 2024-02-21
BL-1849 Add caching forlibkey_articles

nomadicoder:


nomadicoder:
FYI -- This is work in progress.  I has duplicate code, for now I'm looking for feedback as to the use of Rails caches.

dkinzer (approved):

PR #4249, opened 2024-02-26
BL-1840 Make web_concurrency configurable

sensei100:
I will follow up this PR with one to charts to add the environment variable there.

### DIFF
--- a/.docker/app/Dockerfile.prod
+++ b/.docker/app/Dockerfile.prod
@@ -10,12 +10,12 @@ COPY . .
 
 ENV LD_PRELOAD="libjemalloc.so.2" \
       MALLOC_CONF="dirty_decay_ms:1000,narenas:2,background_thread:true" \
-      WEB_CONCURRENCY="8" \
       RUBY_YJIT_ENABLE="1"
 
 USER root
 
 ARG SECRET_KEY_BASE
+ARG WEB_CONCURRENCY
 
 # libc6-compat is required for m1 build.
 RUN apk add -U --no-cache \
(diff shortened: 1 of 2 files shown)


.docker/app/Dockerfile.prod
```diff
@@ -10,12 +10,12 @@ COPY . .
 
 ENV LD_PRELOAD="libjemalloc.so.2" \
       MALLOC_CONF="dirty_decay_ms:1000,narenas:2,background_thread:true" \
-      WEB_CONCURRENCY="8" \
       RUBY_YJIT_ENABLE="1"
 
 USER root
 
 ARG SECRET_KEY_BASE
+ARG WEB_CONCURRENCY
```
dkinzer, line 18: I don't think just a docker build arg alone will work because build args go away after the build and it needs to stay as an system env. I was thinking maybe a combo of a build arg plus the env would do the trick. though, i'm not 100 sure.

sensei100: @dkinzer I added the ENV Variable to the charts deployment template.  Is that not enough?

dkinzer: Unfortunately it's not because the point of this is to make it configurable locally too and the chart change would just manage the deployed change.  Also ARG doesn't work alone because ARG is supposed to be exclusively for branching the build process.  Hmm.. Maybe we can just add --env to run in Makefile and that would be enough for local override?

sensei100: @dkinzer I added it as an env var to the make run command.  I did run it locally and the variable was set to what I expected it to be.

dkinzer (approved):

PR #4259, opened 2024-02-29
BL-1829-update-digital-collections-results-on-bento-page

cdoyle-temple:


ebtoner:
@cdoyle-temple a couple of notes:
- The standard font color is $midnight-blue, not black. 
- I'd recommend putting the cdm styles in _bento.scss rather than its own file.

PR #4273, opened 2024-03-05
BL-1829 Update Digital Collections results on bento page

cdoyle-temple:


ebtoner:
@cdoyle-temple -- looks good. Eventually, I'd love to set this up so that we have a single variable that we rely on for the collections string for both the API and this link, but we can look into that later.

ebtoner (approved):

PR #4298, opened 2024-03-12
BL-1829 update digital collections results on bento page

cdoyle-temple:


ebtoner:
@cdoyle-temple -- This is definitely a step in the right direction, but reading the code, I think that we're still querying https://digital.library.temple.edu/digital/bl/dmwebservices/index.php?q=dmGetCollectionList/json three times, once per result item. Can we update this to do the collection query once per bento search and then refer back to the same json when matching the collection alias to the collection name for each result?

ebtoner:
@cdoyle-temple -- if it's okay with you, I will merge this for now and then follow up on limiting https://digital.library.temple.edu/digital/bl/dmwebservices/index.php?q=dmGetCollectionList/json to one call in a separate PR.

ebtoner (approved):

PR #4333, opened 2024-03-26
Update webpack package versions

sensei100:


sensei100:
Nevermind, looks like the last dependabot PR resolved it.

PR #4344, opened 2024-04-01
Add timeout to CDM HTTP image requests and remove unnecessary image option

ebtoner:
I attempted to add timeout -- please let me know if this is not the right approach. I can't figure out how to test this in rspec, so let me know if you have any ideas for that. 

Also, in looking at the image options more closely, I don't think we want to be using what was defined as the "default_image", so I've removed that.


dkinzer (approved):
this looks good to me.

dkinzer (approved):
this looks good to me.

PR #4440, opened 2024-05-21
BL-1873 Fix top margin on sortAndPerPage div

sensei100:


ebtoner:
@sensei100 -- could we modify https://github.com/tulibraries/tul_cob/blob/df70b45627dacf6736a5afbad90040de800ddc3f/app/views/catalog/_explanation_div.html.erb#L2 instead? I think that we could change mt-3 to my-3 to fix the margin issue and avoid introducing an !important declaration separately into the stylesheets.

sensei100:
@ebtoner Sorry, I thought I removed that.  It doesn't need the important attribute.  I think it makes sense to keep the styling on this div.  Moving it to the explanation div means it might still look off if someone hides that.

ebtoner (approved):

PR #4448, opened 2024-05-24
BL-1866 Create RequestData model for request related data

ebtoner:
This is part of ongoing request refactoring work. Changes include: 
- Creating a RequestData model for anything that prepping/transforming bib item data for the request forms. 
- Moving the majority of methods from CobAlma::Requests (in the lib directory) to the RequestData model, with some changes to the methods themselves
- Moving request_level logic from AlmawsController to the RequestData model 
- Further refactoring AlmawsController, including adding get_bib_request_options

I'm interested in feedback on the model design, so please let me know if you have suggestions!

Things still in progress / to do include refactoring request_options lookup, including holding ids methods still in CobAlma::Requests, and looking at improvements to the request submission code. I'm also thinking that I may not need to define instance variables for everything pulled from @request_data in the controller, but that would require a lot more refactoring of the views that I'm not ready to take on right now.


spec/models/almaws_response_spec.rb
dkinzer, line 1: What's the value in keeping this file around.  If all the functionality that was being tested here is being tested elsewhere now?

ebtoner: @dkinzer -- These particular specs/functionality haven't been moved anywhere. The related model, Blacklight::Alma::Response, is currently inactive, but I decided against deleting it entirely in case it proved useful at a later point. So instead I commented out both the model and this file. 

For context, the inactive model is related to when we were thinking about adding a pagination feature to the item displays, which is not what we currently do. We just use [this method](https://github.com/tulibraries/alma_rb/blob/d1eb8042f4460b4346719f2c40d523874e212b30/lib/alma/bib_item_set.rb#L52) from the alma_rb gem in [get_bib_items](https://github.com/tulibraries/tul_cob/blob/0dfd445650a8b53f893185ab6380833f14606f94/app/controllers/almaws_controller.rb#L218) to grab all of the bib items at once instead. 

If you think it's better to just delete both files to avoid confusion, I can do that. I don't have a strong opinion either way!

PR #4464, opened 2024-05-29
BL-1870 move availability helper methods to methods amd modules

nomadicoder:
DO NOT MERGE this Pull Request.

Here we move methods that do not render HTML to a model concern to lighten the Availability Helper. These methods don't necssasarily need to be in the concern if there is better location for them to go, or if they duplicate methods in a related model.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -629,6 +629,7 @@ GEM
 
 PLATFORMS
   arm64-darwin-21
+  arm64-darwin-23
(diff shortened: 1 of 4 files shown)


ebtoner:
@nomadicoder -- with your permission, i'd like to go ahead and close this PR, since we've worked this together since then.

nomadicoder:
@ebtoner Yes.  We may close this PR.

PR #4466, opened 2024-05-31
BL-1865 Add time estimates to request confirmation message

ebtoner:
- Added new RequestConfirmation model to determine delivery estimate from request submission API response
- Updated message text accordingly (to be reviewed with Justin), removing old helper method in the process
- Added correpsonding spec and fixtures

As always, I welcome any feedback and suggestions to improve this code!

### DIFF
--- a/app/models/request_confirmation.rb
+++ b/app/models/request_confirmation.rb
@@ -18,8 +18,7 @@ class RequestConfirmation
   end
 
   def delivery_estimate
-    unless pickup_location.blank?
-      if pickup_location == "JAPAN" || pickup_location == "ROME"
+      if pickup_location == "JAPAN" || pickup_location == "ROME" || pickup_location.blank?
         nil
       elsif pickup_location == "MAIN" && sent_from == "ASRS"
         "1 hour, delivered from the Charles Library BookBot when open"
@@ -28,7 +27,6 @@ class RequestConfirmation
       else
         "1-3 business days"
       end
-    end
   end
 
   def delivery_estimate?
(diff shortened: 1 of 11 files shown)


ebtoner:
Argh, there's one spec failing that I neglected to test locally. Fixing now.

app/models/request_confirmation.rb
```diff
@@ -0,0 +1,47 @@
...
+    @sent_from = response&.managed_by_library_code
+    @pickup_location = pickup_location
+  end
+
+  def message
+    I18n.t("requests.default_success_message") + delivery_estimate_message + I18n.t("requests.request_status_message")
+  end
+
+  def delivery_estimate
+    unless pickup_location.blank?
```
dkinzer, line 21: So this method can return `nil` in three cases:

`pickup_location.blank?` ||  [ "JAPAN", "ROME"].include(pickup_location)`

I recommend instead of breaking up the return `nil` return value in two statements (the `unless` and the `if`), just fold it into the `if` statement and get rid of the `unless`.

dkinzer:
This looks great. I just have the one suggestion I made.

app/models/request_confirmation.rb
ebtoner: Good call! Just pushed this update

dkinzer (approved):

app/models/request_confirmation.rb
ebtoner: Ack, just realized that I missed how you updated the conditional with .include, which is better. Just pushed that update :)

PR #4489, opened 2024-06-12
Revert "Bl-1848: Add action caching. (#4426)"

dkinzer:
This reverts commit 0dfd445650a8b53f893185ab6380833f14606f94.

### DIFF
--- a/app/controllers/query_list_controller.rb
+++ b/app/controllers/query_list_controller.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class QueryListController < ApplicationController
-  caches_action :show, expires_in: 1.hours, cache_path: Proc.new { |c| c.request.url }
+  caches_action :show, expires_in: 12.hours, cache_path: Proc.new { |c| c.request.url }
 
   def show
     search_service = search_service_class.new(config: blacklight_config, user_params: params)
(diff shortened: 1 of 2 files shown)


app/controllers/query_list_controller.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class QueryListController < ApplicationController
-  caches_action :show, expires_in: 1.hours, cache_path: Proc.new { |c| c.request.url }
+  caches_action :show, expires_in: 12.hours, cache_path: Proc.new { |c| c.request.url }
```
ebtoner, line 4: @dkinzer -- should this be kept at 1 hour? If not, can you explain why we should update this?

dkinzer: This is the revert of the work we did so that it's back to what it was like before we made these changes.  I can put it back to one hour but I don't remember why it as at 12 hours.

ebtoner: Let's keep it at 1 hour. I think that this PR included the decision to make the cache 1 hour, which was applied in more than one place. Do you want to update that in this PR or create a separate one?

ebtoner (approved):

PR #4501, opened 2024-06-19
BL-1806 Files captured in latest security scan

nomadicoder:


.gitlab-ci.yml
```diff
@@ -10,7 +10,7 @@ variables:
   IMAGE: "tulibraries/librarysearch"
   HARBOR: "harbor.k8s.temple.edu"
   HELM_EXPERIMENTAL_OCI: "1"
-  SAST_EXCLUDED_PATHS: "spec, test, tests, tmp, app/controllers/application_controller.rb" # TODO: Remove after BL-1806 completed
+  SAST_EXCLUDED_PATHS: "spec, test, tests, tmp, app/controllers/application_controller.rb, config/environments/development.rb, config/environments/test.rb, app/controllers/concerns/quik_pay.rb, app/controllers/sessions_controller.rb, app/models/blacklight/primo_central/repository.rb, bin, app/helpers/catalog_helper.rb, app/helpers/search_helper.rb, app/helpers/advanced_helper.rb, app/javascript/packs/controllers/explanation_controller.js" # TODO: Remove after BL-1806 completed
```
dkinzer, line 13: I'm fine adding the tests files  to to the exclude path but for code that is actually run in production I am hesitant to approve without knowing the following:

* Why are these files throwing a security alert
* What has been done to fix the problem (other than ignoring it)

nomadicoder: @dkinzer Here are the problems: file:///Users/stevenng/Projects/cob/artifacts-2024-06-19/gl-sast-report%203.json

nomadicoder: @dkinzer I would agree.  I have no problem with this in QA, but we need to address this before going to production.  IIt looks like BL-1806 provides for this discussion.

sensei100: @nomadicoder The only files that I think should be added here would be the ones that are not used in a production environment.  For the others, we need to investigate solutions.

sensei100: @nomadicoder Please remove all routes that are not related to dev/testing from the exclude_paths so that we can merge this. We can address the actual vulnerabilities in a new PR.

.gitlab-ci.yml
```diff
@@ -10,7 +10,7 @@ variables:
   IMAGE: "tulibraries/librarysearch"
   HARBOR: "harbor.k8s.temple.edu"
   HELM_EXPERIMENTAL_OCI: "1"
-  SAST_EXCLUDED_PATHS: "spec, test, tests, tmp, app/controllers/application_controller.rb" # TODO: Remove after BL-1806 completed
+  SAST_EXCLUDED_PATHS: "spec, test, tests, tmp, config/environments/development.rb, config/environments/test.rb # TODO: Remove after BL-1806 completed
```
sensei100, line 13: @nomadicoder Has BL-1806 been completed?  If so, we need to remove the TODO comment.  If not, we need to keep that controller route that the comment refers to.

nomadicoder:
@sensei100 Comment removed.  Please review.

sensei100 (approved):

PR #4506, opened 2024-06-27
BL-1870 Move items_json_display methods to solr document and create new Lookupable module

ebtoner:
Apologies for the large size of this PR! One of the issues with the availability helper methods is that they are used across different controllers, views, and helpers, so a lot of cleanup was needed. 

The availability helper file includes three categories of methods: 1) the actual availability status updates, which use Alma API data 2) business logic related to the solr field "items_json_display" 3) other formatting/rendering related to the physical item views. This PR focuses on the 2nd category. 

Here is a summary of the changes made: 
- Moved methods that transform the Solr field "items_json_display" (or the nested json fields) into the Solr document model
- Created the module "Lookupable" for library_name_from_short_code and location_name_from_short_codes, since these are used both with Solr document items and request data (ex. for pickup locations)
- Refactored the methods "document_availability_info" and "sort_order_for_holdings" into "document_items" and "document_items_grouped" respectively
  - This includes merging the app-generated fields "library", "location", and "call_number_display" into items_json_display (similar to how we handle "availability") for use in the item views.
  - It also transforms the library/location codes using Lookupable on the model side rather than in the view.
- Cleaned up the Solr document model, removing outdated methods related to "materials_data" 
  - This was previously used as the primary Alma API lookup for bib items, I believe, but it's confusing/unnecessary to have this happening here still. 
  - I also updated the email mapping for holdings information, which is the only place we were still using "materials_data". 
- Included Lookupable in request models and removed old helper methods from Almaws controller
  - This included moving the logic for the availability helper method "item_level_library_name" to RequestData, since it is unrelated to availability and uses "library_name_from_short_code."
- Cleaned up related views, helpers, and specs based on these changes 
  - One thing that will need to revisited is the logic/displays related to Aeon. 

As always, happy to discuss any improvements to this code!


nomadicoder (approved):

sensei100:
@ebtoner I like this approach.  Thanks for all the work you have done on this.

PR #4536, opened 2024-07-19
BL-1865 Update delivery estimate message and do some refactoring

ebtoner:


cdoyle-temple (commented):
@ebtoner Should a new class go in its own file, and then be included here?

cdoyle-temple (approved):

PR #4568, opened 2024-08-01
Temporarily disable the semgrep analyzer while Andrew works to fix bu…

dkinzer:
…g in it.

Per Andrew's request: We should disable this analyzer to not be blocked by the false positives that are being caught.  He is fixing the underlying issue.


sensei100 (approved):

dkinzer:
closing this because Andrew now would prefer that we document the false positive to skip issue.

ebtoner:
@dkinzer -- should this not be merged then? That's fine, but I was a bit confused by you repoening the PR after closing.

PR #4595, opened 2024-08-19
Revert to using MD5 hash, but exclude it from SAST scans

sensei100:


dkinzer (approved):

dkinzer:
looks like the rspec test is still using a sha256 value

PR #4661, opened 2024-09-11
BL-1904: Fix error on logout

dkinzer:
A new Rails 7 security feature breaks logouts because devise trys to redirect user to fim.temple.edu in produciton on logoout.

We can monkey patch devise or we can disable the feature.

For now I've disabled the feature.

I tried adding an exception for fim.temple.edu, bu that didn't work.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -47,6 +47,10 @@ module Tulcob
       g.fixture_replacement :factory_bot
     end
     #config.log_level = :debug
+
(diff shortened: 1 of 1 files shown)


ebtoner:
@dkinzer -- What do you mean by disabling the feature?

ebtoner (approved):

sensei100 (approved):

PR #4666, opened 2024-09-12
Update packages that address recent CVEs

nomadicoder:
- Remove waivers
- Update yarn package versions


nomadicoder:
@dkinzer @sensei100 Should I just upgrade the send package and it's related dependencies instead of doing a wholesale yarn upgrade?

nomadicoder:
Attempting to set the version of the send package doesn't seem to work. Though the version is specified in package.json, Trivy scanner doesn't recognize it.

sensei100 (approved):

PR #4670, opened 2024-09-13
Waiver for newly caught CVE-2020-7598 and CVE-2021-44906

nomadicoder:
- Though these are old CVEs, they were just caught for the Minimist package. We await updates to the package.


nomadicoder:
@cdoyle-temple Is there something that needs to be updated with the bootstrap update.  These are pretty old CVE's

nomadicoder:
Remediated by @cdoyle-temple with https://github.com/tulibraries/tul_cob/pull/4676

PR #4692, opened 2024-09-23
BL-1903 Double Primo API timeout time

nomadicoder:


sensei100:
@nomadicoder I left some comments on the Jira ticket for this work. I would rather see us add a configuration for retries and try shorter timeouts with retries.

PR #4759, opened 2024-11-11
Bl-1923 refactor quik pay log output

cdoyle-temple:


app/controllers/concerns/quik_pay.rb
```diff
@@ -23,26 +23,26 @@ def quik_pay_callback
     validate_quik_pay_hash(params.except(:controller, :action))
     validate_quik_pay_timestamp(params["timestamp"])
 
-    log = { type: "alma_pay", user: current_user.id, transactionStatus: params["transActionStatus"] }
+    log = { type: "alma_pay", user: current_user.uid, transactionStatus: params["transActionStatus"], transactionMessage: "Failed Payment" }
```
sensei100, line 26: @cdoyle-temple This will log whether the payment is successful or failed, so it doesn't make sense to hard code the failed payment at the end.

app/controllers/concerns/quik_pay.rb
```diff
@@ -23,26 +23,26 @@ def quik_pay_callback
...
 
-    type, message = do_with_json_logger(log) {
+    if params["transactionStatus"] == "1"
+      balance = Alma::User.send_payment(user_id: current_user.uid);
+      log[:transactionMessage] = "Successful Payment" if balance.paid?
```
sensei100, line 30: @cdoyle-temple Will the transaction message include the balance?

PR #4776, opened 2024-11-19
BL-1923: Fix quikpay error handling.

dkinzer:
Updates quikpay feature to log all user payment information.

Also makes sure that quikpay errors are caught by honeybadger.

Also makes sure  to let user know what to do in case of error.


app/controllers/concerns/quik_pay.rb
```diff
@@ -3,8 +3,22 @@
 module QuikPay
   extend ActiveSupport::Concern
 
+  NO_ACCESSE_MESSAGE = "You do not have access to this pay online feature. If you believe this is incorrect, please call 215-204-8212."
```
sensei100, line 6:
```suggestion
  NO_ACCESS_MESSAGE = "You do not have access to this pay online feature. If you believe this is incorrect, please call 215-204-8212."
```

sensei100 (approved):

PR #4789, opened 2024-12-03
Alma config update

cdoyle-temple:


sensei100:
@cdoyle-temple I did not know that this file existed.  I thought that we were using alma.yml.example for this information.  Can you please remove that file as part of this PR?  It doesn't make sense to have both an example and a a local file with the same information.

sensei100 (approved):

PR #5081, opened 2025-04-18
Bl-1864 update request confirmation message styling

cdoyle-temple:


sensei100 (changes requested):
@cdoyle-temple Please add the comment about removing once we upgrade to blacklight 8.

sensei100 (approved):

PR #5091, opened 2025-04-23
Bl 1868 combine request item from book bot and request item for pickup dropdowns

cdoyle-temple:
This should finish the work of removing the asrs pickup location form.


app/models/solr_document.rb
```diff
@@ -116,8 +116,8 @@ def document_items_grouped
     grouped_items = document_items
       .group_by { |item| item["library"] }
       .transform_values { |library| library.group_by { |item| item["location"] }.sort.to_h }
-    sorted_items = grouped_items
-      .sort_by { |library, locations| [library == "Charles Library" ? 0 : 1, library] }.to_h
+    #sort grouped items by library and location
+    grouped_items.sort_by { |library, locations| [library == "Charles Library" ? 0 : 1, library] }.to_h
```
sensei100, line 120: @cdoyle-temple I'm confused about this change.  I don't think this is related to bookbot requests.  Am I wrong about that?

cdoyle-temple: It came up in the IDE as an unused variable, so the sort_by doesn't need to be assigned to anything as ruby methods return that last thing run/declared, just needs to be run, which it does. Just cleaned it up to be less confusing with no dangling variables.

sensei100: I prefer to isolate code changes to the work that is applicable to the PR.  I will merge it this time, but in the future I would rather have cleanup work be done in a separate PR.

sensei100 (approved):

PR #5172, opened 2025-07-09
creates secure-tmp folders as root

cdoyle-temple:


cdoyle-temple:
@dkinzer This is to get an install folder for ruby/rails to install in the "restricted" environment. I've updated our other rails apps to use this approach, but librarysearch charts were not successfully upgraded yet. It's a two part approach, that does not involve a PVC, where the charts are updated and the app's Dockerfile is also updated to enable the chart's changes. I'm looking into what is different in librarysearch.

PR #5185, opened 2025-07-21
Update cache version

nomadicoder:


nomadicoder:
Changes made in CVE-2025-7339-waiver instead

PR #5309, opened 2025-10-01
Add new cdm collections

tomatoseason:
update cdm_collections_list to display new collections in digitized collections bento


dkinzer (commented):
Looks good.

PR #5645, opened 2026-03-19
BL-2005 Add unit coverage for stale catalog tracking sessions

nomadicoder:
- add controller specs for CatalogTrackingResilience behavior
- verify stale search_session keys are cleared on show
- verify valid search context is preserved on show
- verify track stores search-session params and redirects to record show
- keep stale-tab feature coverage for user-visible no-500 flow


nomadicoder:
Closing as obsolete.  The stale catalog tracking-session implementation and the controller coverage introduced by this branch are already present on main.

PR #5698, opened 2026-04-24
Devo 1367 upgrade rails 8.1

dkinzer:


.docker/ci-app/Dockerfile.ci
```diff
@@ -1,6 +1,7 @@
 FROM ruby:3.4-alpine
 
-ENV RAILS_ENV="test"
+ENV RAILS_ENV="test" \
+    BUNDLER_VERSION="2.5.9"
```
sensei100: @dkinzer Is there a way to get around having to hard-code the bundler version?

.github/workflows/lint-test.yml
```diff
@@ -31,12 +31,6 @@ jobs:
           fetch-depth: 0
           submodules: recursive
 
-      - name: Set up Ruby
```
sensei100, line 34: Why is this being removed?

.github/workflows/lint-test.yml
```diff
@@ -62,11 +56,14 @@ jobs:
           exit 1
 
       - name: Install gems inside container (uses cache tarballs)
-        run: $DC bundle install --jobs 4 --retry 3 --path vendor/bundle
+        run: $DC bundle install --jobs 4 --retry 3
 
       - name: Install Javascript dependencies in container
         run: $DC bundle exec yarn install --frozen-lockfile
 
+      - name: Build stylesheets for test assets
```
sensei100: The other applications still have a precompile step here.  Can we use that to stay consistent?

app/assets/stylesheets/application.scss
```diff
@@ -1,28 +1,29 @@
...
-*/
-
 @import "partials/_vars.scss";
 @import "bootstrap";
+@import "blacklight_advanced_search.css";
+@import "blacklight_range_limit.css.scss";
 @import "selectize.bootstrap5.scss";
 @import "blacklight.scss";
 @import "blacklight_marc";
-@import "partials/*";
```
sensei100, line 26: Are wildcards really no longer supported here?  I worry about people not remembering to add a partial to this list.

app/views/layouts/blacklight.html.erb
```diff
@@ -34,14 +34,25 @@
...
-    <%= opensearch_description_tag application_name, opensearch_catalog_url(:format => "xml") %>
+    <%= opensearch_description_tag application_name, opensearch_catalog_url(format: "xml") %>
     <%= favicon_link_tag %>
     <%= stylesheet_link_tag "application", media: "all" %>
     <%= stylesheet_link_tag "application", "//use.fontawesome.com/releases/v5.6.1/css/all.css" %>
-    <%= javascript_include_tag 'application', 'data-turbolinks-track': 'reload', defer: true %>
-    <%= render partial: "shared/campus_closed" %>
-    <%= javascript_pack_tag "application", 'data-turbolinks-track': 'reload', defer: true %>
     <%= csrf_meta_tags %>
+    <%= javascript_include_tag "jquery/dist/jquery.js", defer: true %>
```
sensei100: Is there a way to get around individual imports here?

public/robots.txt
```diff
@@ -1,104 +1 @@
-# See http://www.robotstxt.org/robotstxt.html for documentation on how to use the robots.txt file
```
sensei100, line 1: Why are we removing this?

docker-compose.ci.yml
```diff
@@ -37,6 +37,8 @@ services:
       ARCHIVESSPACE_PASSWORD: fake-password
       RAILS_ENV: test
       RACK_ENV: test
+      BUNDLER_VERSION: "2.5.9"
```
sensei100: I really would like to avoid hard-coding this.

Gemfile
```diff
@@ -97,7 +98,6 @@ group :test do
   gem "simplecov"
   gem "simplecov-lcov"
   gem "spring"
-  gem "sqlite3", "~> 1.7.3"
```
sensei100, line 100: Are we not using sqlite in testing anymore?

sensei100:
@dkinzer I don't see any updates to the config/application.rb file.  This needs to be updated: config.load_defaults 8.1

sensei100 (changes requested):
See comments

.docker/ci-app/Dockerfile.ci
dkinzer: I couldn't figure out how to.  Initially I had to pin the bundler instance in my bl7.x work.  Maybe that trickled down.  But then on our github actions for tul_cob I was getting nokogiri gem installation issues until I pinned this.

.github/workflows/lint-test.yml
dkinzer: Because we don't actually use it. We run the ruby inside the apline docker image not the ruby we set up outside of it.

.github/workflows/lint-test.yml
dkinzer: I can move this to the Dockerfile.ci too.  Librarysearch is weird in that it uses a Dockrfile.ci ... If we were using Dockerfile.prod we wouldn't need this st\ep because assets build happens on in the Dockerfile.prod.  Other projects like Press-6 just use Docker.prod, do you want me to move to that to keep it consistant?  I don't actually like that we are using Dockerfile.ci in this project.

public/robots.txt
dkinzer: I'll add it back.

Gemfile
dkinzer: I was running into a bundle install issue that removing this helped with.  It get's pulled in from blaacklight so we don't need to add it to our Gemfile unless we wasn to override it.

app/assets/stylesheets/application.scss
dkinzer: Unfortunately dartsass-rails does not support globbing: https://docs.publishing.service.gov.uk/manual/migrate-to-dart-sass-from-libsass.html#other-issues

If you like I can look into another sass compiler that does?

.docker/ci-app/Dockerfile.ci
sensei100: It's just that we are not doing that in any other repositories and I think it will be hard to keep up with it.

.github/workflows/lint-test.yml
sensei100: I'm fine with moving to one Dockerfile.

app/assets/stylesheets/application.scss
sensei100: Okay, I want to use the same thing in all of our projects so we will stick with dart-sass.

public/robots.txt
dkinzer: looks like rails update command got rid ot it. Not sure why it would have done that.

dkinzer: i see it in the public/robots.txt  file so I'm not sure why this says we are removing it.

dkinzer:
@sensei100 I tried to address all your concerns.

.docker/ci-app/Dockerfile.ci
sensei100: Can you try to pin it to the version in our Gemfile.lock file?

sensei100:
@dkinzer I'm still concerned about pinning Bundler and removing sqlite3.

sensei100 (approved):

PR #5764, opened 2026-05-14
BL-2033 add-display-text-to-blank-description-values

cdoyle-temple:


app/models/request_data.rb
```diff
@@ -203,30 +205,38 @@ def remove_by_campus(campus)
           ["GINSBURG"]
         when :PODIATRY
           ["PODIATRY"]
-        when :HARRISBURG
```
sensei100, line 206: @cdoyle-temple Harrisburg still exists.  It was only supposed to be removed from the default pickup locations and I did that work in a different ticket.  Please put these changes back.

sensei100 (changes requested):
See comments

sensei100 (approved):

PR #5808, opened 2026-05-29
Add bot challenge page

dkinzer:


config/initializers/bot_challenge_page.rb
```diff
@@ -0,0 +1,51 @@
...
+   config.skip_when = ->(config) {
+     current_page?(okcomputer_path)
+   }
+
+  # Hook after a bot challenge is presented, for logging or other
+  # config.after_blocked = ->(bot_challenge_controller) {
+  # }
+
+
+  # How long will a challenge success exempt a session from further challenges?
```
sensei100, line 33: @dkinzer Is there a default for this setting if we don't set it here?  If so, what is it?

dkinzer: In the examples in the README it made me think like there is no defaults.

sensei100: If there isn't a default, should we set this?  I think it makes sense that people don't have to keep doing challenges once successful.

dkinzer: @sensei100, sorry I thought your comment was about adding a skip for certain pages.  The default challenge exemption is 24 hours:

https://github.com/samvera-labs/bot_challenge_page/blob/main/lib/bot_challenge_page/config.rb#L33-L34

sensei100 (approved):

PR #5895, opened 2026-07-09
adds p16002coll39 to cdm collection list

tomatoseason:
Adds new collection to the cdm collection list


sensei100:
@tomatoseason I created a Jira ticket for this and am working on refactoring this to a separate file.  This is just an example file, not the real one so adding the collection here won't work.  I think this PR should be closed.

PR #5919, opened 2026-07-21
Add essential packages (6279730)

cdoyle-temple:


sensei100:
@cdoyle-temple This won't be enough to make it pass.  You also need to add files with the CVE name.

sensei100 (changes requested):
See comment

cdoyle-temple:
> @cdoyle-temple This won't be enough to make it pass. You also need to add files with the CVE name.

@sensei100 This is the method that worked in Manifold earlier. It should just work.

https://git.temple.edu/tulibraries/manifold/-/commit/ab0de863eb7dba5d95b26be531e73aeb905d0277

sensei100:
@cdoyle-temple It looks like there is a fix for this already.  Why not just update the gem instead of documenting it?

cdoyle-temple:
@sensei100 Just going on the messages from the scan and the remediation suggested. I'll look to see if a gem update will fix this.

PR #5924, opened 2026-07-22
BL-2054 UI updates for citations

sensei100:


app/views/catalog/_citation.html.erb
```diff
@@ -1,14 +1,13 @@
...
-<% if @response.documents.one? %>
-  <%= render partial: "catalog/single_citation", locals: { document: @response.documents.first } %>
-<% else %>
-  <%= render "catalog/grouped_citations" %>
-<% end %>
-
-<div class="modal-footer justify-content-start ps-0">
-  <p>Remember to check citations for accuracy before including them in your work.</p>
-</div>
+  <div class="citation-warning justify-content-start ps-0" style="color: #E7201D;">
```
cdoyle-temple: There's an inlined color style here that is also declared in the css.

cdoyle-temple (changes requested):
Inlined style should be removed. It'a already in the css for .citation-warning > p

app/views/catalog/_citation.html.erb
sensei100: Thanks!  I took out the inline style.

cdoyle-temple (approved):

PR #5933, opened 2026-07-27
BL-2049 Remove obsolete digital_help_allowed? predicate

nomadicoder:
Remove the obsolete `digital_help_allowed?` following the Digital Help request option removal.

The predicate's only remaining production use was in request-option counting for the request form. Removing it eliminates unused logic and keeps the request-option count aligned with the remaining visible request options.

### DIFF
--- a/spec/helpers/almaws_helper_spec.rb
+++ b/spec/helpers/almaws_helper_spec.rb
@@ -456,6 +456,33 @@ RSpec.describe AlmawsHelper, type: :helper do
       end
     end
 
+    context "only a booking is allowed" do
+      let(:books) {}
+      let(:equipment) {}
+      let(:document) { SolrDocument.new("items_json_display" =>
+        [{ "item_pid" => "23237957740003811",
+        "item_policy" => "5",
+        "permanent_library" => "AMBLER",
+        "permanent_location" => "media",
+        "current_library" => "AMBLER",
+        "current_location" => "media",
+        "call_number" => "DVD 13 A165",
+        "holding_id" => "22237957750003811" }]
+          ) }
+      let(:json) {
+        { request_option:
+          [{
+          "type" => { "value" => "BOOKING", "desc" => "Booking" },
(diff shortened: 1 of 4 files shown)


spec/helpers/almaws_helper_spec.rb
```diff
@@ -456,30 +456,50 @@
...
-        "permanent_location" => "media",
-        "current_library" => "AMBLER",
-        "current_location" => "media",
-        "call_number" => "DVD 13 A165",
-        "holding_id" => "22237957750003811" }]
-          ) }
-      let(:json) {
-        { request_option:
-          [{
-          "type" => { "value" => "BOOKING", "desc" => "Booking" },
```
sensei100, line 475: @nomadicoder I am not sure about this test removal.  I don't think we are removing bookings from the site.

sensei100 (changes requested):
@nomadicoder See the test changes related to hathitrust has me concerned.  We need to make sure the hathitrust links are still working as expected since that integration is not being removed at this time.

spec/helpers/almaws_helper_spec.rb
nomadicoder: Good catch. I intended to add coverage for the remaining digitization-only case after removing `digital_help_allowed?`.  I didn't intend to remove booking coverage. I'll restore the booking example and keep the new digitization example as an additional regression test.

nomadicoder:
@sensei100 - Good point. The removed examples tested digital_help_allowed?, but they also provided indirect coverage of hathitrust_link_allowed?. Since HathiTrust integration is not being removed, I'll check for existing hathitrust_link_allowed? coverage and add it if needed. The digital_help_allowed? tests should remain removed.

nomadicoder:
@sensei100 I checked further. HathiTrust rendering is already covered in `hathitrust_helper_spec.rb`. The removed examples in `solr_document_spec.rb` only exercised the obsolete `digital_help_allowed?` path.

sensei100 (approved):

PR #5934, opened 2026-07-27
BL-2063-update-skip-link-for-skip-to-search

cdoyle-temple:


sensei100 (changes requested):
I think this was what was asked on the ticket, but I think that was a mistake.  If we go straight to the input then a non-sighted person wouldn't know that there were additional filters for searching.  

I think the reason this broke is that the id of the select dropdown changed at some point.  I think we should still skip to the dropdown and just fix the link.

sensei100 (approved):

PR #5942, opened 2026-07-29
BL-669 add controller test coverage for stimulus endpoints

nomadicoder:
Expands user_controller test coverage for stimulus endpoints. Covers success and failing paths.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -28,7 +28,7 @@ class UsersController < ApplicationController
     if holds.success?
       render partial: "users/holds_details", layout: nil, locals: { holds: }
     else
-      render "Problem!"
+      render plain: "Problem!"
     end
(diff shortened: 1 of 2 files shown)


nomadicoder:
While adding controller coverage for the Stimulus endpoints, An existing issue in the error path was reveiled. The loans, holds, and fines actions use:

```
render "Problem!"
```

Rails interprets the string as a template name rather than a plain-text response, resulting in an ActionView::MissingTemplate exception instead of returning a fallback message.  

This behavior appears to predate this PR and is outside the scope of BL-669, whose focus is adding test coverage. I recommend opening a follow-up Jira issue to replace these calls or another appropriate error response.

In the mean time, I made an interim change to

```
render plain: "Problem!"
```

sensei100 (approved):

PR #5970, opened 2026-08-12
BL-2060  Update Japan item pickup locations.

dkinzer:


app/models/request_data.rb
```diff
@@ -47,32 +47,53 @@ def pickup_locations
...
-      end
-      pickup_locations -= removals
-      if (libraries & ["ROME", "JAPAN"]).present?
-        if libraries.size == 1 || libraries.sort == ["JAPAN", "ROME"]
-          pickup_locations = libraries
+    campuses = get_campuses(libraries)
+    removals = lib_removals(libraries)
+
+    pickup_locations = get_valid_pickup_locations(campuses) - removals
+
```
cdoyle-temple, line 69: @dkinzer If available_libraries is empty, this looks like it now returns no pickup locations instead of falling back to the default set: ["MAIN", "AMBLER", "GINSBURG", "PODIATRY"]. Before these changes, we had those cases falling back to the default pickup set but now they produce none.

cdoyle-temple (changes requested):
@dkinzer If available_libraries is empty, this looks like it now returns no pickup locations instead of falling back to the default set: ["MAIN", "AMBLER", "GINSBURG", "PODIATRY"]. Before these changes, we had those cases falling back to the default pickup set but now they produce none.

app/models/request_data.rb
dkinzer: Thanks.  There should be a test for that use case.  So I'll add that too.

cdoyle-temple (approved):

PR #5977, opened 2026-08-20
Relocate bot challenge active record

nomadicoder:
Do not merge.  Draft for Comments and discussion.

Refactor the bot challenge in the catalog_controller in order to reduce CPU load prior to issuing 403 to turnstyle.

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class CatalogController < ApplicationController
+  bot_challenge only: :index, if: -> { bot_challenge? }
+
   caches_page :show
 
   include FacetParamsDedupe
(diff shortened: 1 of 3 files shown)


sensei100:
@nomadicoder Because putting the callback before the includes is slightly unusual, I think we should include a short comment explaining why it needs to stay there.

Also, could we avoid asserting queries.length == 1 here? That feels a little brittle since an unrelated framework/gem change could introduce another harmless query and break the spec. The behavior we care about seems to be that no search/session DB work happens before the challenge, so asserting on the absence of those queries may make this test more durable.

sensei100:
@nomadicoder I really don't think we need to have tests for flipflop/activerecord.  That is a temporary measure and will be removed at some point, which would also require removing these tests.

sensei100 (approved):

PR #5978, opened 2026-08-21
Apply bot challenge to facet requests

nomadicoder:
Apply bot challenge to catalog facet requests

Note: to verify from a browser visit a local library search instance with the `/catalog/facet/creator_facet` path from an incognito browser and from a browser from which you've authenticated to Library Serach.  Incognito session should challenge while an authenticated session should take you to the facet page.

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -28,7 +28,7 @@ class CatalogController < ApplicationController
     end
   end
 
-  bot_challenge only: :index, if: -> { bot_challenge? }
+  bot_challenge only: %i[index facet], if: -> { bot_challenge? }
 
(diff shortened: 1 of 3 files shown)


sensei100 (changes requested):
Could we test this through the normal facet UI before enabling the challenge on facet? bot_challenge_page specifically notes that Blacklight's facet "More" requests are made via JS/XHR and recommends exempting requests with sec-fetch-dest == "empty". Otherwise the fetch may receive the challenge page instead of the facet content. The manual test in the PR description only tests navigating directly to /catalog/facet/..., which doesn't exercise the normal browser flow.

sensei100 (approved):

PR #6031, opened 2026-09-16
disable retries in test-env to uncover vcr's replay error

cdoyle-temple:
- config/initializers/primo.rb sets enable_retries = true with retries = 3, and that applies in every environment.
- Most Primo specs use VCR cassettes that hold a single interaction, and the cassettes match on HTTP method only.
- If something fails after the recorded response is replayed (while parsing, validating, logging, etc.), the Primo gem's with_retry catches the error and sends the request again.
- VCR has nothing left to replay, so it raises VCR::Errors::UnhandledHTTPRequestError, and the original exception is lost.

Disabling the retries should uncover the underlying error


sensei100:
@cdoyle-temple This happens locally too, not just in testing.  David and I worked on resolving it when we were working on the blacklight 9 upgrade.  It is fixed there, so I am going to close this.

PR #6033, opened 2026-09-18
BL-2085-fix-publication-date-text-color

cdoyle-temple:


sensei100:
@cdoyle-temple Do you have any sense of what would have caused that to start happening now? I noticed that file hasn’t been changed in about four years, so I’m curious whether you were able to identify what triggered the issue.

cdoyle-temple:
@sensei100 I put this in the ticket: 

This field, when multi-valued, has a dynamically generated class replacing the standard blacklight-date class, which provides the light blue text color. That text color is only specified in three of the possible class values, so needs to be extended to all. I’ve added the blacklight-date class in addition to the dynamically generate class (derived from field_name) to the display tag, and there are no conflicts in style and applies the expected look uniformly, while retaining the ability to be overridden.

So this was an existing bug for four years, only noticed, or reported now. Nothing needed to change about the code for this to happen. Easy to miss as the black blends in with the dark blue so well and there's just nothing to see.

sensei100 (approved):

